PR #38, opened 2022-01-18
 Remove all SOAP implementation

smisidjan:
deleted all soap implementations in the project
deleted the accordion because of an error -> the fix is coming.


lencodes (approved):
Nog even een ticket aanmaken voor Hamid m.b.t. de Accordion die nu eruit is gehaald.

PR #78, opened 2022-01-28
API implementie voor Endpoint

hghanbari:
Er is een nieuwe API implementie voor Endpoint

### DIFF
--- a/pwa/src/components/endpoints/endpointForm.tsx
+++ b/pwa/src/components/endpoints/endpointForm.tsx
@@ -1,80 +1,68 @@
 import * as React from "react";
-import { GenericInputComponent, Spinner, Alert, SelectInputComponent, Card, Accordion } from "@conductionnl/nl-design-system/lib";
-import { isLoggedIn } from "../../services/auth";
-import { Link } from "gatsby";
-import { navigate } from "gatsby-link";
+import {
+  GenericInputComponent,
+  Spinner,
+  Alert,
+  SelectInputComponent,
+  Card,
+} from "@conductionnl/nl-design-system/lib";
+import {Link} from "gatsby";
+import {navigate} from "gatsby-link";
 import {
   checkValues,
-  removeEmptyObjectValues, retrieveFormArrayAsOArray,
+  removeEmptyObjectValues,
 } from "../utility/inputHandler";
 import FlashMessage from 'react-flash-message';
+import APIService from "../../apiService/apiService";
+import APIContext from "../../apiService/apiContext";
 
 interface EndpointFormProps {
   id: string,
 }
 
-export const EndpointForm:React.FC<EndpointFormProps> = ({ id }) => {
-
-  const [context, setContext] = React.useState(null);
+export const EndpointForm: React.FC<EndpointFormProps> = ({id}) => {
+  const [alert, setAlert] = React.useState<Record<string, string>>(null);
   const [endpoint, setEndpoint] = React.useState<any>(null);
-  const [applications, setApplications] = React.useState<any>(null);
+  const [applications, setApplication] = React.useState<any>(null);
   const [showSpinner, setShowSpinner] = React.useState<boolean>(false);
-  const [alert, setAlert] = React.useState<any>(null);
-  const title:string = (id === "new") ? "Create Endpoint" : "Edit Endpoint"
+  const API: APIService = React.useContext(APIContext)
+  const title: string = id ? "Edit Endpoint" : "Create Endpoint";
 
   React.useEffect(() => {
-    if (typeof window !== "undefined" && context === null) {
-      setContext({
-        adminUrl: process.env.GATSBY_ADMIN_URL,
-      });
-    } else if (isLoggedIn()) {
-      if (id !== "new") {
-        getEndpoint();
-      }
-      getApplications();
-    }
-  }, [context]);
+    handleSetApplications()
+    id && handleSetEndpoints()
+  }, [API, id])
 
-  const getEndpoint = () => {
-    setShowSpinner(true);
-    fetch(`${context.adminUrl}/endpoints/${id}`, {
-      credentials: "include",
-      headers: { "Content-Type": "application/json", 'Authorization': 'Bearer ' + sessionStorage.getItem('jwt') },
-    })
-      .then((response) => response.json())
-      .then((data) => {
-        setShowSpinner(false);
-        setEndpoint(data);
+  const handleSetEndpoints = () => {
+    setShowSpinner(true)
+
+    API.Endpoint.getOne(id)
+      .then((res) => {
+        setEndpoint(res.data)
       })
-      .catch((error) => {
-        setShowSpinner(false);
-        console.error("Error:", error);
-        setAlert(null);
-        setAlert({ type: 'danger', message: error.message });
-      });
+      .catch((err) => {
+        throw new Error('GET Endpoint error: ' + err)
+      })
+      .finally(() => {
+        setShowSpinner(false)
+      })
 
   }
+  const handleSetApplications = () => {
+    setShowSpinner(true)
 
-  const getApplications = () => {
-    setShowSpinner(true);
-    fetch(`${context.adminUrl}/applications`, {
-      credentials: "include",
-      headers: { "Content-Type": "application/json", 'Authorization': 'Bearer ' + sessionStorage.getItem('jwt') },
-    })
-      .then((response) => response.json())
-      .then((data) => {
-        setShowSpinner(false);
-        if (data['hydra:member'] !== undefined && data['hydra:member'].length > 0) {
-          setApplications(data['hydra:member']);
-        }
+    API.Application.getAll()
+      .then((res) => {
+        setApplication(res.data)
       })
-      .catch((error) => {
-        setShowSpinner(false);
-        console.error("Error:", error);
-        setAlert(null);
-        setAlert({ type: 'danger', message: error.message });
-      });
-  };
+      .catch((err) => {
+        throw new Error('GET application error: ' + err)
+      })
+      .finally(() => {
+        setShowSpinner(false)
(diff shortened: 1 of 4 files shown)


lencodes (changes requested):
Good morning @hghanbari,

I don't think you tested your code before requesting this review. Please update this pull request with fixes for all errors and then re-request a review.

If you have any questions, let me know.

Note: this code instantly breaks when going to `http://localhost:8000/endpoints`, in the future I would expect you to catch this **before** requesting a review.

pwa/src/apiService/resources/endpoint.ts
```diff
@@ -0,0 +1,25 @@
...
+export default class Endpoint {
+  private _instance: AxiosInstance
+
+  constructor (_instance: AxiosInstance) {
+    this._instance = _instance
+  }
+
+  public getAll = (): Promise<AxiosResponse> => {
+    return this._instance.get('/endpoints')
+  }
```
lencodes, line 12: This method is not used anywhere, so it can be removed from this file.

pwa/src/components/endpoints/endpointForm.tsx
```diff
@@ -1,238 +1,233 @@
...
-    let method = "POST";
-    if (id !== "new") {
-      url = `${url}/${id}`;
-      method = "PUT";
+
+    if (!id) { // unset id means we're creating a new entry
+      API.Endpoint.create(body)
+        .then((res) => {
+          setEndpoint(res.data)
+          navigate('/endpoints')
```
lencodes, line 92: Check in your merge conflicts if this `navigate` maybe has to be removed (Sarai fixed all the redirects).

lencodes: Note: maybe you have to navigate to `/endpoints/${res.data.id???}` something like that.

pwa/src/components/endpoints/endpointForm.tsx
```diff
@@ -1,238 +1,233 @@
...
-      })
-      .finally(() => {
-        setShowSpinner(false);
-      })
-  };
+    if (id) { // set id means we're updating a existing entry
+      API.Endpoint.update(body, id)
+        .then((res) => {
+          setEndpoint(res.data)
+          navigate('/endpoints')
```
lencodes, line 104: Same goes here.

pwa/src/pages/endpoints/[id].tsx
```diff
@@ -54,7 +55,7 @@ const IndexPage = (props) => {
                 aria-labelledby="overview-tab"
               >
                 <br />
-                <EndpointForm id={props.params.id}/>
+                <EndpointForm {...{id}}/>
```
Comment (lencodes) [inline pwa/src/pages/endpoints/[id].tsx:58]: Nice 🛩️

lencodes (changes requested):

pwa/src/components/endpoints/endpointForm.tsx
```diff
@@ -5,140 +5,116 @@ import {
...
   id: string,
 }
 
 export const EndpointForm: React.FC<EndpointFormProps> = ({id}) => {
-
-  const [context, setContext] = React.useState(null);
+  const [alert, setAlert] = React.useState<Record<string, string>>(null);
   const [endpoint, setEndpoint] = React.useState<any>(null);
-  const [applications, setApplications] = React.useState<any>(null);
+  const [applications, setApplication] = React.useState<any>(null);
```
lencodes, line 27: Applications is plural, so should `setApplications` be. You've changed it to `setApplication`.

pwa/src/components/endpoints/endpointForm.tsx
```diff
@@ -5,140 +5,116 @@ import {
...
-        setEndpoint(data);
-      })
-      .catch((error) => {
-        setShowSpinner(false);
-        console.error("Error:", error);
-        setAlert(null);
-        setAlert({type: 'danger', message: error.message});
-      });
 
+  const handleSetEndpoints = () => {
```
lencodes, line 39: You're setting one endpoint, not multiple. Therefore, this function should be singular: `handletSetEndpoint`.

pwa/src/components/endpoints/endpointForm.tsx
```diff
@@ -5,140 +5,116 @@ import {
...
-    let url = `${context.adminUrl}/endpoints`;
-    let method = "POST";
-    if (id !== "new") {
-      url = `${url}/${id}`;
-      method = "PUT";
+
+    if (!id) { // unset id means we're creating a new entry
+      API.Endpoint.create(body)
+        .then((res) => {
+          setEndpoint(res.data)
```
lencodes, line 93: I don't think this `setEndpoint()` actually does anything here because we're navigating away straight after.

pwa/src/components/endpoints/endpointForm.tsx
```diff
@@ -5,140 +5,116 @@ import {
...
+          navigate('/endpoints/${id}')
```
lencodes, line 94: This `id` variable is not the `id` of the created endpoint, and this `navigate` is therefore not going to the right endpoint. 

The `id` you need to be getting is from the response of the `create()` request.

pwa/src/components/endpoints/endpointForm.tsx
```diff
@@ -5,140 +5,116 @@ import {
...
-      })
-      .finally(() => {
-        setLoadingOverlay(false);
-      })
+
+    if (id) { // set id means we're updating a existing entry
+      API.Endpoint.update(body, id)
+        .then((res) => {
+          setEndpoint(res.data)
+          navigate('/endpoints/${id}')
```
lencodes, line 107: You're already at this endpoint, so this `navigate` is redundant.

pwa/src/components/endpoints/endpointForm.tsx
```diff
@@ -230,47 +207,50 @@ export const EndpointForm: React.FC<EndpointFormProps> = ({id}) => {
...
+                                id={"applicationInput"}
+                                nameOverride={"Applications"}
+                              />
+                            )}
                           </div>
                         </div>
                       </div>
+                      )
                     </div>
                   )}
```
lencodes, line 246: You've done some cleanup in this file, which is good. However, during this, I think you added a `)` too much somewhere. Resulting in this `)` actually rendering to the DOM, see screenshot:

image

lencodes (changes requested):

pwa/src/components/endpoints/endpointForm.tsx
lencodes: This would result in `res.data.id`

pwa/src/components/endpoints/endpointForm.tsx
```diff
@@ -5,140 +5,116 @@ import {
...
+    if (!id) { // unset id means we're creating a new entry
+      API.Endpoint.create(body)
+        .then((res) => {
+          setEndpoint(res.data)
+          navigate('/endpoints/${id}')
+        })
+        .catch((err) => {
+          setAlert({type: 'danger', message: err.message});
+          throw new Error('Create endpoint error: ' + err)
+        })
```
lencodes, line 99: After the `.catch()` we still need a `finally()` to stop the `loadingOverlay`, this can be implemented as followed:

```
.finally(() => { setLoadingOverlay(false) })
```

pwa/src/components/endpoints/endpointForm.tsx
```diff
@@ -5,140 +5,116 @@ import {
...
+    if (id) { // set id means we're updating a existing entry
+      API.Endpoint.update(body, id)
+        .then((res) => {
+          setEndpoint(res.data)
+          navigate('/endpoints/${id}')
+        })
+        .catch((err) => {
+          setAlert({type: 'danger', message: err.message});
+          throw new Error('Update endpoint error: ' + err)
+        })
```
lencodes, line 112: After the .catch() we still need a finally() to stop the loadingOverlay, this can be implemented as followed:

.finally(() => { setLoadingOverlay(false) })

lencodes:
@hghanbari Also, the `EndpointsTable` has not yet been updated.

pwa/src/apiService/resources/endpoint.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export default class Endpoint {
+  private _instance: AxiosInstance
+
+  constructor (_instance: AxiosInstance) {
+    this._instance = _instance
+  }
+  public getAll = (): Promise<AxiosResponse> => {
+    return this._instance.get('/attributes')
+  }
```
lencodes, line 11: You're retrieving `/attributes` and not `/endpoints` here. Resulting in the wrong response.

pwa/src/apiService/resources/endpoint.ts
```diff
@@ -0,0 +1,23 @@
+import { AxiosInstance, AxiosResponse } from "axios"
```
lencodes, line 1: This file does not have consistent whitespace (we're expecting one white line between all functions).

pwa/src/components/endpoints/endpointsTable.tsx
```diff
@@ -1,56 +1,33 @@
...
       })
-      .catch((error) => {
-        setShowSpinner(false);
-        console.log("Error:", error);
-        setAlert(null);
-        setAlert({ type: 'danger', message: error.message });
-      });
-  };
+      .catch((err) => {
+        throw new Error('GET Entities error: ' + err)
```
lencodes, line 23: You're getting Endpoints here, this should reflect in the error message.

lencodes (changes requested):
Hi @hghanbari this code results in errors. This is because in `endpoint.ts` the `getAll()` function is getting the `attributes` and not the `endpoints`.

For some other small changes see the comments.

lencodes (approved):
👍🏼

PR #84, opened 2022-01-31
Alle tabel layout aanpassen

hghanbari:


pwa/src/components/applications/applicationsTable.tsx
```diff
@@ -74,7 +74,7 @@ export default function ApplicationsTable() {
                       headerName: " ",
                       renderCell: (item: { id: string }) => {
                         return (
-                          <Link to={`/applications/${item.id}`}>
+                          <Link className="utrecht-link d-flex justify-content-end" to={`/applications/${item.id}`}>
```
lencodes, line 77: Following feedback counts for all updated `<Link />` elements:

You've added the class `utrecht-link`, why is this? As far as I see this does not actually add any styling that we want to be adding (most is even instantly overwritten by the `Bootstrap` styling).

(See the styling that this class adds to the button):

```
    color: var(--utrecht-link-color, blue);
    text-decoration: var(--utrecht-link-text-decoration, underline);
    text-decoration-skip-ink: all;
    text-decoration-thickness: max(var(--utrecht-link-text-decoration-thickness), 1px);
    text-underline-offset: var(--utrecht-link-text-underline-offset);
```

I think you can remove this class from all the updated elements.

lencodes (changes requested):
Looks good 👍🏼 

You missed the `EntitiesTable`, this edit button does not yet align right.

PR #90, opened 2022-01-31
CON-2258 endpoint log placement

smisidjan:
removed log tab when creating an endpoint


lencodes (changes requested):
Looks good. While testing this, I noticed that the Logs table is also visible when you're creating an endpoint.

Could you maybe fix that as well in this pull request?

(Logs for a not-yet-created endpoint don't exist, so neither should the table).

PR #93, opened 2022-02-01
changed routing forms

smisidjan:
didn't changed data Form because there is a fix in an other pr


pwa/src/components/attributes/attribuutForm.tsx
```diff
@@ -144,25 +156,22 @@ export const AttributeForm: React.FC<AttributeFormProps> = ({ attributeId, entit
...
         })
     }
 
     if (attributeId) { // set id means we're updating a existing entry
       API.Attribute.update(body, attributeId)
         .then((res) => {
-          setAttribute(res.data)
-          navigate(`/entities/${entityId}`)
+          setAttribute(res.data);
+          setShowSpinner(false);
```
lencodes, line 170: The `setShowSpinner(false)` should not happen in the `.then()`, but rather in the `.finally()`.

Reasoning behind that is: when there's an error, the `Spinner` will always stay, but logically it should be removed when the request is finished regardless of the outcome.

lencodes: Please check for all files changed in this pull request

lencodes (approved):
Looks good 👍🏼 

If you can check (and update) the one comment I left, I think it would be a good addition. Other than that, it's approved.

PR #96, opened 2022-02-02
CON-2260 added overlay spinners

remko48:


pwa/src/components/applications/applicationForm.tsx
```diff
@@ -34,6 +34,7 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({id}) => {
   const [alert, setAlert] = React.useState<Record<string, string>>(null);
   const [application, setApplication] = React.useState<IApplication>(null);
   const [showSpinner, setShowSpinner] = React.useState<boolean>(false);
+  const [saveSpinner, setSaveSpinner] = React.useState<boolean>(false);
```
lencodes, line 37: This feels a bit duplicated, consider the following two suggestions:

1. We are now using the `saveSpinner` and `showSpinner` separately from each other, it might be better to completely remove the `showSpinner` and just always use the overlay (you can maybe check this with @rubenvdlinde )
2. The name `saveSpinner` is too specific, what if we want an overlay spinner when we're deleting something, fetching something, or any other reason. Therefore, we should rename this (to e.g. `OverlaySpinner` or `LoadingOverlay`).

pwa/src/components/applications/applicationForm.tsx
```diff
@@ -147,6 +148,13 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({id}) => {
                         <Spinner/>
                       ) : (
                         <div>
+                          {saveSpinner === true ? (
+                            <div className="spinner-overlay">
+                              <div className="spinner-overlay-content">
+                                <Spinner />
+                              </div>
+                            </div>
+                          ) : ( <div></div>)}
```
lencodes, line 157: Instead of using the syntax:

```
{ saveSpinner === true ? <></> : <></> }
```

You can use:

```
{ saveSpinner && <></> }
```

This has a couple of benefits:

1. Less code
2. Better readability
3. You are no longer rendering an empty element (the `<div></div>`) when the `saveSpinner` state is false.

Also, for future reference, when you're checking if a boolean is true, you do not have to explicitly say so. Resulting in: 

```
if (saveSpinner === true)
if (saveSpinner)
```

pwa/src/components/attributes/attribuutForm.tsx
```diff
@@ -214,6 +215,13 @@ export const AttributeForm: React.FC<AttributeFormProps> = ({attributeId, entity
                         <Spinner/>
                       ) : (
                         <div>
+                          {saveSpinner === true ? (
+                            <div className="spinner-overlay">
+                              <div className="spinner-overlay-content">
+                                <Spinner />
+                              </div>
+                            </div>
```
lencodes, line 223: Right now, we are creating the following boilerplate every time we want to include this component:

```
<div className="spinner-overlay">
  <div className="spinner-overlay-content">
    <Spinner /> // only this line might not be considered boilerplate
  </div>
</div>
```

I would say that this piece of code is a componenten of itself. Right now, when we need to make changes to this component, we would have to make changes everywhere we've included it like this.

A better way of structuring this component would be to:

1. Make a folder in de `/components` folder (named something like `loadingOverlay`, depending on how you're going to call your component)
2. Create your component here with the `.tsx` and `.css` files (see feedback on styling to understand why we need a `.css` file here)
3. Lastly, instead of writing the boilerplate code everywhere, it can become something like:

```
{ saveSpinner && <LoadingOverlay /> } // this is your current state name
{ loading && <LoadingOverlay /> } // i would suggest something like this
```

pwa/src/styles/main.css
```diff
@@ -219,3 +219,24 @@ a:hover {
     border: none !important;
     background: inherit !important;
 }
+
```
lencodes, line 222: All the required styling can go to the created `.css` file mentioned earlier (should be located somewhere like `/components/loadingOverlay/loadingOverlay.css`).

To actually use this styling within your new component, you can include it like so (this code goes in `loadingOverlay.tsx`):

`import "./loadingOverlay.css";`

And to scope the `.css` code you can use another prefix (you're using `spinner-overlay` now, which is fine but can be more specific):

```
.loadingOverlay {
  ...
}

.loadingOverlay > .loadingOverlay-content {
  ...
}

.loadingOverlay > ... // always prefix with .loadingOverlay so that you can never style another component by accident
```

pwa/src/styles/main.css
```diff
@@ -219,3 +219,24 @@ a:hover {
...
+.spinner-overlay {
+    height: 100%;
+    width: 100%;
+    display: block;
+    position: fixed;
+    z-index: 1;
+    top: 0;
+    left: 0;
+    background-color: rgb(0,0,0);
```
lencodes, line 231: This line is redundant I think.

pwa/src/styles/main.css
```diff
@@ -219,3 +219,24 @@ a:hover {
...
+    background-color: rgba(0,0,0, .3);
+}
+
+.spinner-overlay-content {
+    position: relative;
+    top: 25%;
+    width: 100%;
+    text-align: center;
+    margin-top: 30px;
+}
```
lencodes, line 241: This is not centered right now, and is using multiple `top` variables. 

A easy fix for this would be to do the following:

1. Remove any and all styling from `.spinner-overlay-content` (therefore you can also remove the class itself in the component)
2. Add the following styles to `.spinner-overlay`:

```
display: flex; // so not block
justify-content: center; // makes it center horizontally
align-items: center; // makes it center vertically
```

lencodes: [CSS Flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) is a widely accepted new standard for positioning elements (and is now also supported by all major browsers). Take some time to read some documentation, we will be using it a lot 😄

lencodes (changes requested):
Hey @remko48, the end product is already looking very good! I see some things we can update so that it will be a bit more robust. If you have any questions, let me know.

**Out of scope for this pull request, but** interesting to know anyway: the way we've created the overlay now could create some issues due to the `z-index` dependency. Right now, we're saying that the overlay has a `z-index: 1;`. What happens when someone, somewhere else in the DOM, says something has `z-index: 2;`? (Depending on its location in the DOM) it will probably be rendering over the overlay. To fix this you can use [React Portals](https://reactjs.org/docs/portals.html), this renders elements outside of the DOM-structure, giving you the ability to always be on top.

pwa/src/components/applications/applicationForm.tsx
```diff
@@ -149,6 +151,7 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({id}) => {
                     <Spinner/>
                   ) : (
                     <div>
+                          {overlaySpinner && <LoadingOverlay /> }
```
lencodes, line 154: This is the easiest spot to point this out: it might be best to choose a consistent name for both the `state` and the `component`. You're now using `overlaySpinner` as `state` and `LoadingOverlay` as `component` name.

This would be much more clear when we do one of the following:

1. Change the `state` variable in all files to `loadingOverlay` (and its setter to `setLoadingOverlay`) or
2. Change the `component` name to `OverlaySpinner`.

That way, you get a much more clearer line here (and in the other files, like e.g.):

`{ loadingOverlay && setLoadingOverlay }`

Another `state` variable that would make sense would just be `loading` and `setLoading` so your syntax could become `loading && <LoadingOverlay />`.

Doesn't really matter which one you choose, but I think implementing one of these suggestions would be a good addition 👍🏼

pwa/src/components/attributes/attribuutForm.tsx
```diff
@@ -216,6 +218,7 @@ export const AttributeForm: React.FC<AttributeFormProps> = ({attributeId, entity
                     <Spinner/>
                   ) : (
                     <div>
+                          {overlaySpinner && <LoadingOverlay /> }
```
lencodes, line 221: This indentation, and the same line in all files, is not correct.

lencodes (changes requested):
Looks much better!

Two really small comments, after that you can merge it!

PR #99, opened 2022-02-02
Source formulier refactor

hghanbari:


pwa/src/components/common/multiDimensionalArrayInput.tsx
```diff
@@ -59,7 +59,7 @@ export function MultiDimensionalArrayInput(props: MultiDimensionalArrayInputProp
       </div>
       <br />
       <div className="separator-solid" />
-      <h5>Add {props.label ?? props.id}</h5>
+      <h5>Create {props.label ?? props.id}</h5>
```
lencodes, line 62: 👍🏼

lencodes (changes requested):
Hey @hghanbari, the changed code looks good. However, I do have one concern.

The `MultiDimensionalArrayInput` used to output its contents as such:

image

But now, after saving a source, the header shows as such:

image

So the value "Key" is overwritten to "0". I'm aware we spoke about this issue before, but I did not expect this issue to come up in this pull request because this is still using the old way of working.

Please check and fix before merging anything.

PR #100, opened 2022-02-03
Template Type in de handler form selectbox 

hghanbari:
Ik heb deze form code schone gemaakt .

### DIFF
--- a/pwa/src/components/handlers/handlerForm.tsx
+++ b/pwa/src/components/handlers/handlerForm.tsx
@@ -137,8 +137,7 @@ export default function HandlerForm({id, endpointId}) {
         ? event.target.entity.value : null,
       template: event.target.template.value
         ? event.target.template.value : null,
-      templateType: event.target.templateType.value
-        ? event.target.templateType.value : null,
+      templateType: event.target.templateType.value,
       conditions,
       skeletonIn,
       skeletonOut,
@@ -254,13 +253,31 @@ export default function HandlerForm({id, endpointId}) {
                           />
                         </div>
                         <div className="col-6">
-                          <GenericInputComponent
-                            type={"text"}
-                            name={"templateType"}
-                            id={"templateTypeInput"}
-                            data={handler && handler.templateType && handler.templateType}
-                            nameOverride={"Template Type"}
-                          />
+                          {handler !== null && handler.templateType !== null ? (
+                            <SelectInputComponent
+                              options={[{name: "twig", value: "twig"},
+                                {name: "markdown", value: "markdown"},
+                                {name: "username-password", value: "username-password"}
+                              ]}
+                              name={"templateType"}
+                              id={"templateTypeInput"}
+                              nameOverride={"Template Type"}
+                              required={true}
+                              data={handler.templateType}
+                            />
+                          ) : (
+                            <SelectInputComponent
+                              options={[
+                                {name: "twig", value: "twig"},
+                                {name: "markdown", value: "markdown"},
+                                {name: "restructuredText", value: "restructuredText"}
+                              ]}
+                              name={"templateType"}
+                              id={"templateTypeInput"}
+                              nameOverride={"Template Type"}
+                              required={true}
+                            />
+                          )}
                         </div>
                       </div>
                       <br/>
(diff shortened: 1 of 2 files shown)


pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -193,201 +193,243 @@ export default function HandlerForm({id, endpointId}) {
...
+                        <div className="col-6">
+                          {handler !== null && handler.type !== null ? (
+                            <SelectInputComponent
+                              options={[
+                                {name: "twig", value: "twig"},
+                                {name: "markdown", value: "markdown"},
+                                {name: "restructuredText", value: "restructuredText"},
+                              ]}
+                              name={"templateType"}
+                              id={"templateTypeInput"}
```
lencodes, line 300: On line `138` you're selecting `even.target.templateType.value` and not `event.target.templateTypeInput`. This selector is based on this `id` you're setting.

So you're setting `templateTypeInput`, but trying to retrieve the data from `templateType` which results in an error (see screenshot).

image

pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -193,201 +193,243 @@ export default function HandlerForm({id, endpointId}) {
...
+                              options={[
+                                {name: "twig", value: "twig"},
+                                {name: "markdown", value: "markdown"},
+                                {name: "restructuredText", value: "restructuredText"},
+                              ]}
+                              name={"type"}
+                              id={"templateTypeInput"}
+                              nameOverride={"Template Type"}
+                            />
+                          )}
```
lencodes, line 315: This entire block can be simplified I think.

Basically, we're checking if we have data for the select element and based on that we render another select element. However, we could be rendering the same select element in both cases, and only populate it with data when we have it.

So, at the top of this block we're doing the following check: `{handler !== null && handler.type !== null ? (`. This entire check can be removed (and therefore we only need the code within the `:` and we can remove the block within the `?`.

Finally, because we've removed the check. We do not know if we have the data to populate the field. Therefore, we have to rewrite this line: `data={handler.templateType}` to `data={handler?.templateType}` the `?` after the variable `handler` means `if (handler) { handler.data }`.

So above means: if there's no handler, we just leave data empty. If there is a handler, we fill the data with handler.templateType.

lencodes (changes requested):
I get an error when creating a new Handler. Check inline comments.

Also, there are some merge conflicts!

pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -138,8 +138,7 @@ export default function HandlerForm({id, endpointId}) {
       template: event.target.template.value
         ? event.target.template.value : null,
       templateType: event.target.templateType.value
-        ? event.target.templateType.value : null,
-      conditions,
+        ? event.target.entity.value : null,
```
lencodes, line 141: You're doing a couple of things in this line:

1. You're setting the `templateType` key within the `body` object (seems good);
2. You're checking if `event.target.templateType.value` (seems good);
3. If the value exists you're setting `event.target.ENTITY.value` (**does not seem right)** and finally
4. If the value did not exist you're setting it to `null` (seems good).

Currently, you're filling the key with irrelevant data, and because of that the it doesn't behave as intended.

pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -254,13 +253,18 @@ export default function HandlerForm({id, endpointId}) {
...
+                              options={[
+                                {name: "twig", value: "twig"},
+                                {name: "markdown", value: "markdown"},
+                                {name: "restructuredText", value: "restructuredText"}
+                              ]}
+                              name={"type"}
+                              id={"templateType"}
+                              nameOverride={"Template Type"}
+                              required={true}
+                              data={handler?.templateType}
```
lencodes, line 266: Something's going wrong here, the `SelectInputComponent` doesn't actually show the right value (it's always on the first option "twig").

lencodes (changes requested):
Hey @hghanbari. I feel as if you could've figured out that this code doesn't work **before creating this pull request**. 

When updating a `Handler` it's quite clear that the data is not being saved (the `select` input does not actually show the updated value on reload).

So once again, please test your work before submitting it for a review.

PR #101, opened 2022-02-03
Feature/con 2266/welcome modal

bbrands02:


pwa/src/components/welcomeModal/welcomeModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+import APIContext from "../../apiService/apiContext";
+import { Modal } from "@conductionnl/nl-design-system";
+
+const WelcomeModal: React.FC = () => {
+  const API: APIService = React.useContext(APIContext);
+  const buttonRef = React.useRef(null);
+
+  React.useEffect(() => {
+    // Check route
+    if (window.location.href.split("/").at(-1) !== 'applications' || (window.location.href.split("/").at(-1) !== 'new' && window.location.href.split("/").at(-2) !== 'applications')) {
```
lencodes, line 14: I think this if statement got more complex than necessary, the following does the same:

`window.location.pathname.includes('applications') // results in a true or false depending if 'applications' is in the path`

Above in mind, I think the following set-up might be a bit better for readability and in a more consistent React way of working:

```
...
  const [applicationsCount, setApplicationsCount] = React.useState<number>(null)
  const buttonRef = React.useRef(null);

  React.useEffect(() => { handleSetApplicationsCount() }, [API])

  React.useEffect(() => {
    if (applicationsCount === 0 || !buttonRef.current || window.location.pathname.includes('applications')) return

    buttonRef.current.click()
  }, [buttonRef])

  const handleSetApplicationsCount = (): void => {
    API.Application.getAll()
      .then((res) => { setApplicationsCount(res.data.length) })
      .catch((err) => { throw new Error('GET applications error: ' + err) })
  }
...
```

pwa/src/components/welcomeModal/welcomeModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <p>It seems you haven't created a application yet, would you like to
+          <a
+            className="utrecht--link"
+            data-bs-dismiss="modal"
+            aria-label="Close"
+            onClick={() => { navigate('/applications/new') }}>
+            {` create an application `}
+          </a>
+          first or you can close this modal if you want to find things out yourself. </p>
+      </>
```
lencodes, line 38: Generally speaking, a call to action is always written in an _active_ manner. This message is quite passive.

Current: "would you like to create an application ... "

Target (e.g.): "Click here to create your first application"

Also, the "or you can close this modal if you want to find things out yourself." might not be the best addition because the modal will pop-up on every screen. Maybe just be more direct and say that a user **has to** create an application to get started.

lencodes (changes requested):
Looks good. Couple of small comments.

pwa/src/components/common/layout.tsx
```diff
@@ -26,26 +27,27 @@ export default function Layout({ children, title = "", subtext = "" }) {
...
             </div>
-          </> : <Login />
-        }
+          </div>
+          <WelcomeModal />
+        </> : <Login />
+      }
 
-      </APIProvider>
+    </APIProvider>
```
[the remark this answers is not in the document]
bbrands02: This is because of the 2 line indentation I configured earlier in my IDE. So this should be good?

PR #104, opened 2022-02-03
Feature/con 2195/info overlay

MWest2020:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes/
| BC breaks?    | 
| Deprecations? | 
| Tests pass?   | 
| Fixed tickets | #2195
| License       | MIT
| Doc PR        | 

Implemented the requested modal for all components with a help button


pwa/src/components/applications/applicationsTable.tsx
```diff
@@ -74,7 +100,7 @@ export default function ApplicationsTable() {
                       headerName: " ",
                       renderCell: (item: { id: string }) => {
                         return (
-                          <Link className="utrecht-link d-flex justify-content-end" to={`/applications/${item.id}`}>
+                          <Link to={`/applications/${item.id}`}>
```
lencodes, line 103: Why is this change necessary? 

There's at least one thing breaking due to this change, the _Edit_ button was aligned right in its table with the classes `d-flex justify-content-end`.

pwa/src/components/configurations_export_button/ConfigurationsExportButton.tsx
```diff
@@ -39,7 +41,7 @@ export default function ConfigurationsExportButton() {
       onClick={handleExport}
       disabled={loading}
     >
-      {loading ? "Perparing your download...": "Export Configuration"}
+      {loading ? "Preparing your download..." : "Export Configuration"}
```
lencodes, line 44: Nice 😄

pwa/src/components/endpoints/endpointsTable.tsx
```diff
@@ -31,105 +41,139 @@ export default function EndpointsTable() {
...
-              <i className="fas fa-sync-alt mr-1" />
-              <span className="mr-2">Refresh</span>
-            </a>
-            <Link to="/endpoints/new">
-              <button className="utrecht-button utrecht-button-sm btn-sm btn-success">
-                <i className="fas fa-plus mr-2" />
-                Create
+  React.useEffect(() => {
+    // na apiService refactor
+    // handleSetApplications();
```
lencodes, line 61: Remove comments.

pwa/src/components/entities/entitiesTable.tsx
```diff
@@ -1,29 +1,48 @@
...
-    setShowSpinner(true)
+    setShowSpinner(true);
     API.Entity.getAll()
       .then((res) => {
-        setEntities(res.data)
+        setEntities(res.data);
       })
       .catch((err) => {
-        throw new Error ('GET Applications error: ' + err)
+        throw new Error("GET Applications error: " + err);
```
lencodes, line 30: GET Entities error*

pwa/src/components/sources/sourcesTable.tsx
```diff
@@ -1,36 +1,67 @@
...
-    setShowSpinner(true)
+    setShowSpinner(true);
     API.Source.getAll()
       .then((res) => {
-        setSources(res.data)
+        setSources(res.data);
       })
       .catch((err) => {
-        throw new Error ('GET Applications error: ' + err)
+        throw new Error("GET Applications error: " + err);
```
lencodes, line 30: Get Sources error*

pwa/src/styles/main.css
```diff
@@ -219,3 +219,8 @@ a:hover {
     border: none !important;
     background: inherit !important;
 }
+
+#helpModal {
+    text-align: left;
+    color: black;
+}
```
lencodes, line 226: Usually we reserve using an `id` (`#`-selector) for edge case javascript selecting (it was very popular in things like `jQuery` and is still in use in things like `Bootstrap` so that's why we need to actually add the `#helpModal` id at this moment).

In React however, we pretty much never want to add an `id` to a component. Therefore, we also would like to not use the `#` selector in our `css`.

Above being said, usually we would have a unique `className` for each component and we can scope our styling like that. However, because of the way everything's structured right now, that would be a bit of a rework.

Therefore: I would like to suggest to just swap this `#helpModal` with `.modal` 👍🏼

lencodes (changes requested):
Looks good! Couple of small comments.

MWest2020 (commented):
Please review changes so I don't merge something and break dev branch

PR #106, opened 2022-02-03
favicon added

bbrands02:


pwa/src/components/common/layout.tsx
```diff
@@ -8,6 +8,7 @@ import { APIProvider } from "../../apiService/apiContext";
 import APIService from "../../apiService/apiService";
 import { isLoggedIn } from "../../services/auth";
 import Login from "../../pages/login";
+import favicon from "../../images/conduction_logo_blauw.svg";
```
lencodes, line 11: This is currently giving a Typescript error. The reason is that this file does not actually export any module. A `svg` won't be starting to export a model either, so we need to fix this somehow.

The solution is to create a `custom.d.ts` file in the `/src` folder with the following contents:

```
declare module "*.svg" {
  const content: any;
  export default content;
}
```

Now, we have created a module export for `.svg` files, however we still need to accept this module into our Typescript config, like so:

```
// This file is located at /pwa/tsconfig.json
{
  "compilerOptions": {
    "target": "ESNext",
    "module": "ESNext",
    "lib": ["dom"],
    "jsx": "preserve",
    "moduleResolution": "node",
    "allowJs": true,
    "outDir": "./build",
  },
  "include": ["src/custom.d.ts", "src/components"] // only this line is new, including the just created file
}
```

lencodes: With above comment implemented, the error should go away.

Note: if it doesn't you might need to restart your typescript server (on VSCode you can do this by pressing cmd+shift+p and search for restart ts server).

pwa/src/components/common/layout.tsx
```diff
@@ -26,26 +27,29 @@ export default function Layout({ children, title = "", subtext = "" }) {
...
+              <Footer />
             </div>
-          </> : <Login />
-        }
+          </div>
+        </> : <Login />
+      }
 
-      </APIProvider>
+    </APIProvider>
```
lencodes, line 53: This should be indented, like it was before.

lencodes (changes requested):

PR #108, opened 2022-02-04
TranslationForm refactor and fix

bbrands02:


lencodes (changes requested):
Hey @bbrands02 the code changes I see, seem very good. However, before I do the actual review, could you please check the following:

1. When I go to edit a `Translation` in the table on `http://localhost:8000/translations`, it redirects me to `/translations/${tables.name}`, resulting in an error. (I'd assume it goes to `../id` and not `../name`);
2. I assumed that this logic would be in `translationTable`. However, it is found in `tableNamesTable` why are those two files separate? Maybe one can be removed? I'm not sure maybe I'm missing something, but please do check.

Let me know when the errors are gone and I'll re-review this!

PR #114, opened 2022-02-08
TextareaComponent

hghanbari:


pwa/src/components/attributes/attribuutForm.tsx
```diff
@@ -22,6 +22,7 @@ import ElementCreationNew from "../common/elementCreationNew";
 import APIService from "../../apiService/apiService";
 import APIContext from "../../apiService/apiContext";
 import LoadingOverlay from "../loadingOverlay/loadingOverlay";
+import {TextareaGroup} from "../common/textareaGroup";
```
lencodes, line 25: The ticket is regarding **Endpoint** forms and not this form. So I think this implementation is not in the right spot.

I understand the reasoning behind the change, **please check with @rubenvdlinde if this change is wanted as we did not refine it.**

pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -20,6 +20,7 @@ import {MultiDimensionalArrayInput} from "../common/multiDimensionalArrayInput";
 import ElementCreationNew from "../common/elementCreationNew";
 import {navigate} from "gatsby-link";
 import LoadingOverlay from "../loadingOverlay/loadingOverlay";
+import {TextareaGroup} from "../common/textareaGroup";
```
lencodes, line 23: The ticket is regarding **Endpoint** forms and not this form. So I think this implementation is not in the right spot.

I understand the reasoning behind the change, **please check with @rubenvdlinde if this change is wanted as we did not refine it.**

pwa/src/components/common/textareaGroup.tsx
```diff
@@ -0,0 +1,39 @@
+import * as _ from "lodash";
```
lencodes, line 1: Do we still want to include this in our `NL-Design-System` repository? 

This would mean:

1. Remove this code from this repository;
2. Add this code to the `NL-Design-System` repository;
3. Create a pull request there and eventually merge it (this merge will give you a new version number of the package);
4. Include this new version number in `package.json` within this repository;
5. Finally, include the TextareaGroup components from the package and no longer from the local file.

PR #115, opened 2022-02-08
CON-2239 Refactor Alert replacement

smisidjan:
* installed gatsby layout packaged
* created headerContext.ts
* created header.tsx
* added header and headerProvider to the layout
* changed the pages and added the setHeader

At the detail page object the title is not shown properly yet, I will make a separate issue for this: GW-24


pwa/src/components/applications/applicationForm.tsx
```diff
@@ -32,14 +32,16 @@ interface ApplicationFormProps {
...
 
 export const ApplicationForm: React.FC<ApplicationFormProps> = ({id}) => {
-  const [alert, setAlert] = React.useState<Record<string, string>>(null);
+  const [_, setAlert] = React.useContext(AlertContext);
   const [application, setApplication] = React.useState<IApplication>(null);
   const [showSpinner, setShowSpinner] = React.useState<boolean>(false);
   const [loadingOverlay, setLoadingOverlay] = React.useState<boolean>(false);
   const API: APIService = React.useContext(APIContext)
   const title: string = id ? "Edit Application" : "Create Application";
+  const [header, setHeader] = React.useContext(HeaderContext);
```
lencodes, line 41: This can be `[_, setHeader]`, because you never read the value.

pwa/src/components/applications/applicationForm.tsx
```diff
@@ -51,6 +53,7 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({id}) => {
         setApplication(res.data)
       })
       .catch((err) => {
+        setAlert({message: err, type: 'danger'})
```
lencodes, line 56: Works like a charm!

I think for the error messages, we want to send two pieces of data to the user:

1. An generic message like "Oops something went wrong" and
2. The actual error message.

Right now, we're only pushing the error message to the user. Could you create a JIRA issue for us to update this error Alert in the future?

pwa/src/components/common/header.tsx
```diff
@@ -1,25 +1,25 @@
 import * as React from "react";
```
lencodes, line 1: Nice!

pwa/src/components/common/header.tsx
```diff
@@ -1,25 +1,25 @@
...
+import {HeaderContext} from "../../context/headerContext";
 
+const Header = () => {
+  const [header, _] = React.useContext(HeaderContext)
```
lencodes, line 5: Because the unused part of this `state` (`setHeader`, which you're saving as `_` right now) is the second value in the array, you can just skip it altogether. Like so: 

`const [header] = React.useContext(HeaderContext)`.

pwa/src/components/common/layout.tsx
```diff
@@ -3,53 +3,59 @@ import Footer from './../footer/footer'
...
-                <Footer />
               </div>
-            </div>
             <WelcomeModal />
-          </> : <Login />
-        }
+            </> : <Login />
+          }
+        </HeaderProvider>
+      </AlertProvider>
```
lencodes, line 58: The indentation is off in this file.

pwa/src/components/common/layout.tsx
```diff
@@ -3,53 +3,59 @@ import Footer from './../footer/footer'
...
   React.useEffect(() => {
     !API && setAPI(new APIService(sessionStorage.getItem('jwt')))
   }, [API])
 
   return (
     API &&
-      <APIProvider value={API}>
+    <APIProvider value={API}>
+      <AlertProvider value={[alert, setAlert]}>
+        <HeaderProvider value={[header, setHeader]}>
```
lencodes, line 35: We're chaining multiple ContextProviders here. Which is still manageable here because it's only three. However, when we need more and more in our Context, this won't work anymore.

Therefore, we will need to create a more generic place where we can chain these ContextProviders and then within the layout we only have to wrap it like (e.g.) `<ContextProvider value={context}>`.

Could you create a JIRA issue for this so we can do this in the future?

lencodes, line 35: Very nice solution like this 👍🏼

pwa/src/components/endpoints/endpointForm.tsx
```diff
@@ -121,156 +118,147 @@ export const EndpointForm: React.FC<EndpointFormProps> = ({id}) => {
       .then((data) => {
         setLoadingOverlay(false);
         setEndpoint(data)
-        method === 'POST' && navigate("/endpoints")
+        method === 'POST' && navigate("/endpoints") && setAlert({message: 'Saved endpoint', type: 'success'})
```
lencodes, line 121: Why are we chaining everything with `&&` here? I think it might be better to keep them on separate lines.

pwa/src/components/entities/entitiesTable.tsx
```diff
@@ -3,13 +3,19 @@ import {Table, Card, Spinner} from "@conductionnl/nl-design-system/lib";
...
   const [entities, setEntities] = React.useState(null);
   const [showSpinner, setShowSpinner] = React.useState<boolean>(false);
   const API: APIService = React.useContext(APIContext)
+  const [_, setAlert] = React.useContext(AlertContext)
+  const [header, setHeader] = React.useContext(HeaderContext);
 
-  React.useEffect(() => { handleSetEntities() }, [API])
+  React.useEffect(() => {
+    setHeader({title: 'Objects', subText: 'An overview of your objects'})
+    handleSetEntities() }, [API])
```
lencodes, line 18: The `}, [API])` should be on a separate line.

pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -173,11 +171,14 @@ export default function HandlerForm({id, endpointId}) {
       .then((response) => response.json())
       .then((data) => {
         setHandler(data)
-        method === 'POST' && navigate(`/endpoints/${endpointId}`)
+        method === 'POST' && navigate(`/endpoints/${endpointId}`) && setAlert({
+          message: 'Saved handler',
+          type: 'success'
+        })
+        setAlert({message: 'Updated handler', type: 'success'})
```
lencodes, line 178: I think, to be more consistent with the rest of our codebase, we should be checking on if `id` was set or not. Instead of this `=== 'POST'` check. 

If `id` is set, it was a `Update`, otherwise it was a `Create`.

Might even be something like:

```
setAlert({
  message: '${id ? "Updated" : "Created"} handler',
  type: ...
})
```

lencodes: Note: above syntax will automatically no longer be needed when we implement the `ApiService` here (because we've decoupled the `create` and `update` functions).

pwa/src/components/handlers/handlerTable.tsx
```diff
@@ -40,19 +39,12 @@ export default function HandlerTable({ id }) {
       .catch((error) => {
         setShowSpinner(false);
         console.log("Error:", error);
```
lencodes, line 41: We should never `console.log` anything in our codebase.

This should be a new error: `Throw new Error (...)`

pwa/src/context/headerContext.ts
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react'
+
+export interface HeaderProps {
+  title: string,
+  subText: string,
+}
+
+export const HeaderContext = React.createContext<[HeaderProps, (value: HeaderProps) => void]>(null)
+export const HeaderProvider = HeaderContext.Provider
```
lencodes, line 9: Nice! 😄

pwa/src/pages/applications/[id].tsx
```diff
@@ -6,15 +6,15 @@ const IndexPage = (props) => {
   const id: string = props.params.id === "new" ? null : props.params.id
 
   return (
```
Comment (lencodes) [inline pwa/src/pages/applications/[id].tsx:8]: You missed some comments here.

pwa/src/pages/applications/index.tsx
```diff
@@ -1,21 +1,20 @@
 import * as React from "react"
-import Layout from "../../components/common/layout";
 import ApplicationsTable from "../../components/applications/applicationsTable";
+import {HeaderContext} from "../../context/headerContext";
 
 const IndexPage = () => {
+  const [header, setHeader] = React.useContext(HeaderContext);
```
lencodes, line 6: I don't think you're actually using these variables.

pwa/src/pages/entities/index.tsx
```diff
@@ -1,10 +1,9 @@
 import * as React from "react";
-import Layout from "../../components/common/layout";
 import EntitiesTable from "../../components/entities/entitiesTable";
 
 const IndexPage = () => {
   return (
-    <Layout title={"Object"} subtext={"An overview of your objects"}>
+    <>
       <title>Gateway - Objects</title>
```
lencodes, line 7: What does this actually do? I don't see it anywhere.

Also, this is not included in the other similar files. We might be able to just delete it and then also delete the empty `<></>` wrapper because `<main>` would wrap around everything.

pwa/src/pages/users/[id].tsx
```diff
@@ -1,16 +1,13 @@
 import * as React from "react";
```
Comment (lencodes) [inline pwa/src/pages/users/[id].tsx:1]: This is just an empty file 😂 

I don't think we have anything built around users. Might be worth to just remove this entire folder altogether.

However, I do understand that that's probably out of scope for this pull request. Maybe you can create a JIRA issue for this?

pwa/src/pages/users/index.tsx
```diff
@@ -1,21 +1,19 @@
...
-        <div className="row">
-          <div className="col-12">
-            <div className="page-top-item">
-              <UsersTable/>
-            </div>
+    <main>
+      <div className="row">
+        <div className="col-12">
+          <div className="page-top-item">
+            <UsersTable/>
```
lencodes, line 10: This doesn't seem to be working for me. Maybe, together with the other comment about users, it might be worth to add this to the JIRA issue as well.

lencodes (changes requested):
Very nice changes! I left quite a few comments (as expected with such a big pull request).

**Note: **many comments are made in just one file, but they do refer to all the similar files (many changes were duplicated, I did not duplicate my comments).

**Note: **check your merge conflicts before merging anything!

lencodes (changes requested):
Hey @smisidjan, a couple of small things, after that, you can merge this:

1. The `EndpointForm` is rendering double (maybe something went wrong during the merge conflicts);
2. The following resources are missing their `Alert` messages: `Objects` (children of `Object Types`), `Translations` and `Handlers` and lastly
3. I left some comments with the question if you could create JIRA issues for some future changes, I can't find them (if you did create them you can just ignore this third comment 👍🏼 )

PR #117, opened 2022-02-09
GW-20 Handle redirect after jwt token expires 

smisidjan:
* added check for redirecting


pwa/src/components/common/layout.tsx
```diff
@@ -22,9 +23,26 @@ import favicon from "../../images/conduction_logo_blauw.svg";
 export default function Layout({ children, title = "", subtext = "" }) {
   const [API, setAPI] = React.useState<APIService>(null)
 
+  const parseJwt = (token) => {
+    try {
+      return JSON.parse(atob(token.split(".")[1]));
```
lencodes, line 28: The use of `atob` is advised against. My IDE gives the following reason: 

```
This function is only provided for compatibility with legacy web platform APIs and should never be used in new code, because they use strings to represent binary data and predate the introduction of typed arrays in JavaScript. For code running using Node.js APIs, converting between base64-encoded strings and binary data should be performed using Buffer.from(str, 'base64') andbuf.toString('base64').
```

Might be worth to check if we can use another alternative.

pwa/src/components/common/layout.tsx
```diff
@@ -22,9 +23,26 @@ import favicon from "../../images/conduction_logo_blauw.svg";
...
+    } catch (e) {
+      return null;
+    }
+  };
+  const decodedJwt = parseJwt(sessionStorage.getItem('jwt'));
+
   React.useEffect(() => {
+    if (decodedJwt === null) {
```
lencodes, line 36: `if (!decodedJwt)` is the same as `if (decodedJwt === null)`.

pwa/src/components/common/layout.tsx
```diff
@@ -22,9 +23,26 @@ import favicon from "../../images/conduction_logo_blauw.svg";
...
+      logout()
+      navigate("/");
+    } else if (decodedJwt.exp * 1000 < Date.now()) {
```
lencodes, line 39: I would stay away from `else if` statements as much as possible. In my opinion, this could just be another if statement, like so:

`if (Date.now() >= decodedJwt?.exp * 1000)` Note: see the ? after decodedJwt, this firstly does the check if `decodedJwt` is available at all.

lencodes (changes requested):
**Edit from the future: check out this pull request: https://github.com/ConductionNL/commonground-gateway-frontend/pull/134. You can probably hook into something here to add the logic.**

Hey @smisidjan this solution looks good to me. I left a couple of small syntax comments below. 

One slightly bigger change I think is necessary (where I think 1. is important and 2. is optional):

1. This _logic_  should not be in `Layout`, we do need to check in `Layout` because this is a generic place which every request will hit (see change 2. for an even better placement of this logic). However, we could (e.g.) make a function in `services/auth` called `validateSession()` and put all this logic in there. This way, `Layout` only needs to have that one line of code in itself. All other logic (the checks) and behaviour (the redirect, etc.) can be within the `services/auth:validateSession()`;

2. (This one is depending on how much time we have, but would be a great addition). Usually, because we want to run this logic on every request, we would use middleware. Middleware is a piece of code that you can put in the _middle_ of blocks of logic. So, in this case, we can put it in the middle of every request. Gatsby has middleware options of itself, [found here](https://www.gatsbyjs.com/docs/reference/functions/middleware-and-helpers/).


**Edit: ** I now see that you have a build error, this happens because you're trying to access `sessionStorage.getItem('jwt')` before it is actually set (on line 33 in layout.tsx). This has to do with Gatsby and React lifecycles. A solution for this would be to access it **within** the `react.useEffect()` (just like on line 44).

pwa/src/services/auth.tsx
```diff
@@ -20,4 +23,14 @@ export const isLoggedIn = () => {
 export const logout = () => {
   setUser({});
   window.sessionStorage.removeItem("jwt");
+  window.sessionStorage.removeItem("user");
 };
+
+export const validateSession = (jwt) => {
+  const decodedJwt = jwtDecode<JwtPayload>(jwt);
+
+  if (decodedJwt?.exp >= Date.now()) {
```
lencodes, line 32: Can `decodedJwt` ever be unset (and therefore optional) here? 

If so, what _should_ happen when it's unset? I'd assume it means there's no valid `jwtToken` and the session is therefore **not validated**. 

I think currently, when the `decodedJwt` is unset, the session would be validated as valid and nothing would happen.

pwa/src/services/auth.tsx
```diff
@@ -20,4 +23,14 @@ export const isLoggedIn = () => {
...
+    logout()
+    navigate("/login");
```
lencodes, line 34: I think navigating back to `/login` should not be part of a `validateSession` method, since it has nothing to do with validating the session itself. 

Even calling the `logout` method is debatable here. It might make more sense to implement one of the following two changes:

1. Let `validateSession` return a boolean (`true` for valid and `false` for invalid), when calling upon the `validateSession` method you can call the `logout` method depending on the outcome or
2. Move `navigate("/login")` to the `logout` method itself.

lencodes (approved):
Approved (please do update the small comments I left). 👍🏼

PR #118, opened 2022-02-09
response time refactor

smisidjan:
* added function millisecondsToSeconds
* changed headerName to Response time
* added renderCell to return the response time in ms and s
* added function to response time in the modal
* updated dummy data


pwa/src/components/logs/logTable/logTable.tsx
```diff
@@ -93,8 +97,13 @@ export const LogTable: React.FC<LogTableProps> = ({ entityId }) => {
                           field: "requestMethod",
                         },
                         {
-                          headerName: "Response time (seconds)",
+                          headerName: "Response time",
                           field: "responseTime",
+                          renderCell: () => {
+                            return (
+                              `${log?.responseTime.toString()}ms (${millisecondsToSeconds(log?.responseTime).toString()}s)`
```
lencodes, line 104: What would happen when `log` does not exist here?

I think it will be printing "mss" because the "ms" and "s" are not optional with the `log`.

Maybe something like `log && '${log.responseTime}ms (${msToSeconds(log.responseTime)})'`

Note that I also removed the .toString here, I think it works fine.

pwa/src/components/logs/logTable/logTable.tsx
```diff
@@ -183,7 +192,7 @@ export const LogTable: React.FC<LogTableProps> = ({ entityId }) => {
                       </tr>
                       <tr>
                         <th>Response time</th>
-                        <td>{log?.responseTime.toString() + ' seconds'}</td>
+                        <td>{millisecondsToSeconds(log?.responseTime).toString() + ' seconds'}</td>
```
lencodes, line 195: I think this one should get the same structure as the other one (so 2342ms (2s))

lencodes (changes requested):

lencodes (changes requested):
Nice 👍🏼 

Only thing I see is that the formatting in both Tables are different from each other and not correct in either of them.

It should be `0000ms (0s)`.

Current `DashboardLogTable` is rendering: `0000ms (0)s`

Current `LogTable` is rendering: `0000ms (0)`

PR #119, opened 2022-02-09
added prettier

smisidjan:
* added package prettier

when deploying, the other developers still have to install the plugin


pwa/package.json
```diff
@@ -36,6 +36,7 @@
   },
   "devDependencies": {
     "dotenv": "^10.0.0",
-    "gatsby-link": "^4.3.0"
+    "gatsby-link": "^4.3.0",
```
lencodes, line 39: I don't think `gatsby-link` is part of this pull request.

pwa/package.json
```diff
@@ -36,6 +36,7 @@
...
+    "prettier": "2.5.1"
```
lencodes, line 40: Do we actually need to install this package?

For Visual Code, we do not. We have an plugin we can install via the plugin hub and than we can just insert the path to this prettier file you've created.

However, maybe for PhpStorm this is necessary? Maybe you can double check.

lencodes (approved):
Top!

PR #121, opened 2022-02-09
GW-8 added icons to buttons

smisidjan:
* added + icon to the add button -> elementCreationNew
* added trash icon to the delete button -> elementCreationNew
* added eye icon to the view button -> logTable


lencodes (changes requested):
Hey @smisidjan, functionally this works fine. However, technically we can include `FontAwesome` icons the _React way_ by doing the following:

1. Import the `FontAwesomeIcon` wrapper and import any required icons:

```
import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
import { faTrash } from "@fortawesome/free-solid-svg-icons";
```

2. Implement the icon where needed (e.g. for the delete button):

```
<FontAwesomeIcon icon={faTrash} /> Delete
```

This way, we do not actually have to manually add the `font-aweomse` classes.

Note: I see you've also used a `mr-2` class, this is a [Bootstrap spacing class](https://getbootstrap.com/docs/4.0/utilities/spacing/), which we don't want to use anymore. If spacing is required, we will have to write our own css to do so.

pwa/src/components/common/multiDimensionalArrayInput.tsx
```diff
@@ -49,7 +51,7 @@ export function MultiDimensionalArrayInput(props: MultiDimensionalArrayInputProp
                       type="button"
                       className="utrecht-button utrecht-button-sm btn-sm btn-danger"
                     >
-                      Delete
+                      <FontAwesomeIcon icon={faTrash} /> Delete
```
lencodes, line 54: This one isn't actually working (at least not in the `AttributeForm`). This is because this button is actually rendered from `pwa/src/components/common/elementCreation.tsx`.

lencodes (approved):
Looks great. Approved (please do check the one comment I left). 👍🏼

pwa/src/components/common/multiDimensionalArrayInput.tsx
smisidjan: Schermafbeelding 2022-02-22 om 17 30 27

PR #122, opened 2022-02-09
GW-18 Align add button

smisidjan:
* refactor columns elementCreation + added new className to elementCreationNew.tsx
* added style for the new className

### DIFF
--- a/pwa/src/components/common/elementCreationNew.tsx
+++ b/pwa/src/components/common/elementCreationNew.tsx
@@ -77,32 +77,42 @@ const ElementCreationNew: React.FC<ElementCreationNewProps> = ({id, label, data,
 
       <div className="row">
         <h5>Create {label}</h5>
-        <div className="col-10">
           {
             select ? (
-              <div className="form-group">
-                <SelectInputComponent
-                  options={options !== null && options.length > 0 ? options : []}
-                  name={selectName} id={`${selectName}Input`}
-                  nameOverride={label}
-                  onChange={(e) => setValue(e.target.value)}
-                />
-              </div>
+              <>
+                <div className="col-10">
+                  <SelectInputComponent
+                    options={options !== null && options.length > 0 ? options : []}
+                    name={selectName} id={`${selectName}Input`}
+                    nameOverride={label}
+                    onChange={(e) => setValue(e.target.value)}
+                  />
+                </div>
+                <div className="col-2 select-elementCreation-button">
+                  <button className="utrecht-button utrecht-button-sm btn-success" onClick={handleAdd} disabled={!value}>Add
+                  </button>
+                </div>
+              </>
             ) : (
-              <input
-                type="text"
-                value={value}
-                placeholder={`Add ${label}`}
-                onChange={(e) => setValue(e.target.value)}
-                className="utrecht-textbox utrecht-textbox--html-input mb-2"
-              />
+              <>
+                <div className="col-10">
+                  <input
+                    type="text"
+                    value={value}
+                    placeholder={`Add ${label}`}
+                    onChange={(e) => setValue(e.target.value)}
+                    className="utrecht-textbox utrecht-textbox--html-input mb-2"
+                  />
+                </div>
+                <div className="col-2">
+                  <div className="input-group">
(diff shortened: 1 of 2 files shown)


pwa/src/components/common/elementCreationNew.tsx
```diff
@@ -77,32 +77,42 @@ const ElementCreationNew: React.FC<ElementCreationNewProps> = ({id, label, data,
 
       <div className="row">
         <h5>Create {label}</h5>
-        <div className="col-10">
           {
             select ? (
```
lencodes, line 81: I think the styling issues started when this `select` option was added to this `ElementCreation` component. I think after the MVP we will be refactoring a lot, so for now we can leave this as is.

However, I think we shouldn't have added this to this component. It makes it more complex and hard to maintain and also hard to build upon.

lencodes: Note: this could've been its own component. 👍🏼

pwa/src/components/common/elementCreationNew.tsx
```diff
@@ -77,32 +77,42 @@ const ElementCreationNew: React.FC<ElementCreationNewProps> = ({id, label, data,
...
-                  nameOverride={label}
-                  onChange={(e) => setValue(e.target.value)}
-                />
-              </div>
+              <>
+                <div className="col-10">
+                  <SelectInputComponent
+                    options={options !== null && options.length > 0 ? options : []}
+                    name={selectName} id={`${selectName}Input`}
+                    nameOverride={label}
```
lencodes, line 87: This label pushes the `select` down, and that's how the margin is created. That being said, in the other implementation of `ElementCreation` we opted to do it without a label. Which I would suggest here as well, because this would be the third label, which seems excessive.

image

pwa/src/styles/main.css
```diff
@@ -227,3 +227,7 @@ a:hover {
 .utrecht-button:disabled:hover {
     cursor: not-allowed;
 }
+
+.select-elementCreation-button {
+    margin-top: 28px;
+}
```
lencodes, line 233: Simply by removing the `label` and removing these `css` rules. The layout would look like this:

image

Also, this way of positioning something can be dangerous. You're positioning the element from within itself, based on outside context (e.g. the button has to go down, because the outside context - in this case the select field - is lower than itself so it doesn't align). What would happen when we want to include this same button somewhere else, but in another context? It would get an unintentional `margin-top`, most likely breaking layout. Besides wanting to re-use this button, what would happen when the context changes? 

E.g.: the padding of the label changes, than the following happens:

image

In this case, we can just remove the label, so it doesn't really matter. In other cases however, we could have wrapped the `select` and `input` components in a `div` and set the position of those elements **within** that `div`, removing outside context from the elements itself.

lencodes (changes requested):

pwa/src/components/utility/selectInput.tsx
```diff
@@ -2,45 +2,52 @@ import * as _ from "lodash";
...
         </select>
       </div>
-    </>
   );
 };
 
-SelectInputComponent.defaultProps = {
+SelectInputForMultiSelect.defaultProps = {
   required: false
 };
```
lencodes, line 53: I don't think this is necessary (when `required` is optional and not set when implemented it will be [Falsy](https://frontend.turing.edu/lessons/module-1/js-truthy-falsy-expressions.html#:~:text=and%20Falsy%20Values.-,Truthy%20and%20Falsy%20Values,NaN,-All%20other%20values)).

pwa/src/components/utility/selectInput.tsx
```diff
@@ -2,45 +2,52 @@ import * as _ from "lodash";
...
+  disabled?: boolean;
 }
 
+
 /**
  * This component generates a select input.
  * @returns Jsx of the generated form.
  */
-export const SelectInputComponent = (props: SelectInputProps) => {
+export const SelectInputForMultiSelect: React.FC<SelectInputProps> = ({
```
lencodes, line 21: When the file is called `SelectInput.tsx`, the component should be called `SelectInput`.

PR #123, opened 2022-02-09
feature translations refactor

bbrands02:


pwa/src/components/translations/tableNamesTable.tsx
```diff
@@ -96,10 +96,10 @@ export default function TableNamesTable() {
                         headerName: " ",
                         renderCell: (tables: { name: string }) => {
                           return (
-                            <Link className="utrecht-link d-flex justify-content-end" to={`/translations/${tables.name}`}>
-                              <button className="utrecht-button btn-sm btn-success">
-                                <i className="fas fa-edit pr-1" />
-                                Edit
+                            <Link className="utrecht-link d-flex justify-content-end" to={`/translation-tables/${tables.name}`}>
```
lencodes, line 99: I think we should refactor this to `/translations-tables/id`. For the following reasons:

1. To be more consistent (we are currently using the `id` for every route in very resource;
2. To be more precise (e.g. right now, when you have a translation _table_ called `new` we will go to the create instead of the edit)

pwa/src/components/translations/translationForm.tsx
```diff
@@ -2,96 +2,95 @@ import * as React from "react";
...
-import {navigate} from "gatsby-link";
+import { Link } from "gatsby";
 import LoadingOverlay from "../loadingOverlay/loadingOverlay";
+import APIService from "../../apiService/apiService";
+import APIContext from "../../apiService/apiContext";
 
 interface TranslationFormProps {
-  id: string,
+  id?: string,
+  tableName?: string
```
lencodes, line 15: I think with these changes, we actually have two forms right now. 

1. Creating a new Translation Table and
2. Creating a new Translation

Now, I understand, that both are actually creating the exact same thing in the background. But because we have such a separation in the frontend, I think we should split the forms as well.

So we get a `TranslationTableForm` and a `TranslationForm`.

I'll give you a quick call on how we could maybe choose to keep them in the same file, but still separate them.

pwa/src/components/translations/translationForm.tsx
```diff
@@ -2,96 +2,95 @@ import * as React from "react";
...
 
-export const TranslationForm: React.FC<TranslationFormProps> = ({id}) => {
-  const [context, setContext] = React.useState(null);
+export const TranslationForm: React.FC<TranslationFormProps> = ({ id, tableName }) => {
   const [showSpinner, setShowSpinner] = React.useState<boolean>(false);
   const [loadingOverlay, setLoadingOverlay] = React.useState<boolean>(false);
-  const [alert, setAlert] = React.useState<any>(null);
   const [translation, setTranslation] = React.useState<any>(null);
-  const title: string = (id === "new") ? "Create Translation" : "Edit Translation"
+  const title: string = (tableName && tableName !== 'new' && !id) ? "Create translation" : (tableName && id) ? "Edit translation" : (tableName && tableName === 'new') ? "Create table" : null
```
lencodes, line 22: By separating the two forms, we wouldn't need complex (and very hard to manage and build upon) checks like these.

pwa/src/components/translations/translationForm.tsx
```diff
@@ -2,96 +2,95 @@ import * as React from "react";
...
-        </FlashMessage>
-      }
       <form id="dataForm" onSubmit={saveTranslation}>
         <Card
           title={title}
           cardHeader={function () {
             return (
               <div>
-                <Link className="utrecht-link" to={"/translations"}>
+                <Link className="utrecht-link" to={translation?.translationTable ? `/translation-tables/${translation.translationTable}` : (tableName && tableName !== 'new') ? `/translation-tables/${tableName}` : '/translation-tables'}>
```
lencodes, line 84: By separating the two forms, we wouldn't need complex (and very hard to manage and build upon) checks like these.

pwa/src/components/translations/translationForm.tsx
```diff
@@ -100,54 +99,60 @@ export const TranslationForm: React.FC<TranslationFormProps> = ({id}) => {
...
+                                  id={"translationTableInput"}
+                                  data={translation && translation.translationTable && translation.translationTable}
+                                  nameOverride={"Name"}
+                                  required />
+                              </div>
+                            </div>
                           </div>
-                        </div>
-                      </div>
+                          <h4 className="utrecht-heading-4 utrecht-heading-4--distanced">You need to create at least one translation when creating a new table</h4>
```
lencodes, line 122: Two notes on this:

1. This heading should be smaller than the Table Title (e.g. _Create Table_) because it's lower in the hierarchy, also, it should be black
2. The ticket mentions a subtitle as well, which is missing.

pwa/src/pages/translation-tables/[table].tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from "react";
+import Layout from "../../components/common/layout";
+import TranslationTable from "../../components/translations/translationTable";
+import TranslationForm from "../../components/translations/translationForm";
+
+const IndexPage = (props) => {
+  const tableName: string = props.params.id === "new" ? null : props.params.table
```
Comment (lencodes) [inline pwa/src/pages/translation-tables/[table].tsx:7]: You're doing three checks within this file if `tableName !== "new"` **after this declaration**.

It can never be "new" right? Only if the table name itself is "new"? Because you're setting it to `null`.

lencodes (changes requested):
Looking very good already. Do have some feedback, see inline comments.

pwa/src/components/translations/translationTable.tsx
```diff
@@ -22,7 +22,7 @@ export default function TranslationTable({id}) {
 
   const getTranslations = (context) => {
     setShowSpinner(true);
-    fetch(`${context.adminUrl}/translations?translationTable=${id}`, {
+    fetch(`${context.adminUrl}/translations?translationTable=${tableName}`, {
```
lencodes, line 25: This is still using the old fetch.

pwa/src/pages/translation-tables/[table].tsx
Comment (lencodes) [reply pwa/src/pages/translation-tables/[table].tsx:7]: Another not: `props.params.id` doesn't exist at all here.

pwa/src/apiService/resources/translation.ts
```diff
@@ -0,0 +1,33 @@
...
+export default class Translation {
+  private _instance: AxiosInstance
+
+  constructor (_instance: AxiosInstance) {
+    this._instance = _instance
+  }
+
+  public getAll = (): Promise<AxiosResponse> => {
+    return this._instance.get('/translations')
+  }
```
lencodes, line 12: This is unused. Therefore, I would recommend removing it.

pwa/src/components/translations/table.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from "react";
+// import "../../style/table.css";
```
lencodes, line 2: You missed a comment (I don't think the file exists, so this comment should just be removed).

pwa/src/components/translations/table.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+interface TableProps {
+  columns: Array<Partial<Record<"field" | "headerName" | "renderCell" | "hidden" | "valueFormatter", any>>>;
+  rows: Array<Record<any, any>>;
```
lencodes, line 6: Why are we using such complex types (an `Array` that holds a `Partial` that holds a `Record`). Is this really necessary? Also, the use of `any` really does beat the purpose most of the time.

Especially the `rows: Array<Record<any, any>>;` doesn't say too much since we're passing `any` twice.

lencodes: But besides the `any` argument, couldn't we make this a much easier type and achieve the same outcome?

pwa/src/components/translations/table.tsx
```diff
@@ -0,0 +1,42 @@
...
+            <th key={index}>{item.headerName ?? item.field}</th>
+          ))}
+        </tr>
+      </thead>
+      <tbody>
+        {rows.map((row, index) => (
+          <tr key={index}>
+            {columns.map((column, idx) =>
+              Object.keys(row).includes(column.field) && !column.hidden ? (
+                column.renderCell ? (
```
lencodes, line 24: By using simpler types, we might be able to simplify this as well.

pwa/src/components/translations/tableNamesTable.tsx
```diff
@@ -1,168 +1,110 @@
...
-} from "@conductionnl/nl-design-system/lib";
-import { isLoggedIn } from "../../services/auth";
 import { Link } from "gatsby";
-import FlashMessage from "react-flash-message";
+import { Table, Card, Spinner, Alert } from "@conductionnl/nl-design-system/lib";
 import APIService from "../../apiService/apiService";
 import APIContext from "../../apiService/apiContext";
 
 export default function TableNamesTable() {
   const [documentation, setDocumentation] = React.useState<string>(null);
```
lencodes, line 8: This is unused.

pwa/src/components/translations/tableNamesTable.tsx
```diff
@@ -1,168 +1,110 @@
...
   const [tableNames, setTableNames] = React.useState<Array<any>>(null);
-  const [context, setContext] = React.useState(null);
   const [showSpinner, setShowSpinner] = React.useState<boolean>(false);
   const [alert, setAlert] = React.useState(null);
```
lencodes, line 11: This is unused.

pwa/src/components/translations/tableNamesTable.tsx
```diff
@@ -1,168 +1,110 @@
...
-        setShowSpinner(false);
-        setTableNames(convertedArray);
+    API.Translation.getTableNames()
+      .then((res) => { 
+        let names = [];
+        let results = res.data.results;
+        for (let i = 0; i < results.length; i++) {
+          names.push({name: results[i]})
+        }
+        setTableNames(names);
```
lencodes, line 27: Generally speaking, when you're trying to map an array into new data, you could best use the `.map()` method (instead of a `for` loop). Like so,

```
const names = res.data.results.map((name) => { 
  return { name: name }
});
```

And since `name` is also the key, we can even simplify `name: name` to `{ ... {name} }`, like so:

`const names = res.data.results.map((name) => { return { ... { name } }})`

The result is the exact same, but without the `for` loop 👍🏼

lencodes (changes requested):
Looks good, left a few small notes.

Also, I mentioned the heading in the form in my other review, this doesn't seem to be updated (the _You need to..._ should be smaller and black):

image

pwa/src/components/translations/translationForm.tsx
lencodes: 1. This heading is now the exact same size as its parent heading (this is now a `h5` and the table title is a `h4`). However, they both have a font-size of `20`. The `h5` should be smaller (e.g. `font-size: 18px`);
2. You're still missing the subtitle.

pwa/src/apiService/resources/translation.ts
```diff
@@ -1,29 +1,33 @@
...
+    return this._instance.put(`/translations/${id}`, JSON.stringify(data))
+  }
 
   public delete = (id: string): Promise<AxiosResponse> => {
     return this._instance.delete(`/translations/${id}`)
   }
+
+  public getTableNames = (): Promise<AxiosResponse> => {
+    return this._instance.get("/table_names")
+  }
```
lencodes, line 32: Most changes here are just formatting changes, these happen because you're not using the `.prettierrc` file. Please install the `Prettier` plugin and start using this file. Then, please re-save all these files so that the formatting is the same as the rest of the team.

pwa/src/components/translations/translationTableForm/translationTableForm.tsx
```diff
@@ -0,0 +1,113 @@
...
+                              id={"translationTableInput"}
+                              data={translation && translation.translationTable && translation.translationTable}
+                              nameOverride={"Name"}
+                              required />
+                          </div>
+                        </div>
+                      </div>
+
+                      <hr />
+                      <h5 className="utrecht-heading-5 utrecht-heading-5--distanced h5-translationForm mb-3">You need to create at least one translation when creating a new table</h5>
```
lencodes, line 100: 1. This heading is now the exact same size as its parent heading (this is now a `h5` and the table title is a `h4`). However, they both have a font-size of `20`. The `h5` should be smaller (e.g. `font-size: 18px`);
2. You're still missing the subtitle.

pwa/src/components/translations/tableNamesTable.tsx
```diff
@@ -1,100 +1,61 @@
...
-        setTableNames(convertedArray);
-      })
-      .catch((err) => {
-        setAlert({ message: err, type: "danger" });
-        throw new Error("GET table names error: " + err);
+    API.Translation.getTableNames()
+      .then((res) => { 
+        const names = res.data.results.map((name) => { 
+          return { name: name }
+        });
```
lencodes, line 22: When using a `.map()` function, the shorthand way of returning the items is by **not adding** the `{}`, like so:

```
const names = res.data.results.map((name) => { name: name } );
```
_or_
```
const names = res.data.results.map((name) => { ...{ name } });
```

**Note: ** I can imagine this being confusing. The `{}` here are not the brackets of opening a function but the brackets of opening the object we're returning 😛

pwa/src/components/translations/translationTable.tsx
```diff
@@ -20,19 +13,17 @@ export default function TranslationTable({ tableName }) {
...
-
-  React.useEffect(() => {
-    setHeader({ title: "Translations", subText: "An overview of your translation objects" });
-  }, [setHeader]);
-
+  
   React.useEffect(() => {
-    handleSetDocumentation(); // we added this
+    setHeader({title: 'Translations', subText: 'An overview of your translations'});
   });
```
lencodes, line 19: This goes for **everywhere** you're setting `setHeader()`:

1. By setting `setHeader` you're triggering a re-render every time and are therefore getting max-depth errors (huge performance issue, we definitely have to fix this)

To fix this, we can add the `setHeader` setter to the `useEffect` dependency array. This means: only re-run this block of code when (most likely if) `setHeader` changes, like so:

```
React.useEffect(() => {
  setHeader({ title: "Translation tables", subText: "An overview of your translation tables" });
}, [setHeader]); // this array is the dependency array
```

2. You're setting the header `setHeader()` in components **and** pages, which means you're setting them double (I'm not sure if you're doing this everywhere but for sure in the following example). I think it's best to set the headers in `pages` and not in `components` (and I think that's how Sarai meant it to be implemented when building it). Example of where I think you're duplicating the set header code (I assume you can just remove it from the component):

`TranslationTable` component:
```
React.useEffect(() => {
  setHeader({ title: "Translations", subText: "An overview of your translations" });
}, [setHeader]);
```

`Translations` page:
```
React.useEffect(() => {
  setHeader({ title: "Translations", subText: "An overview of your translations" });
}, [setHeader]);
```

pwa/src/components/translations/translationTableForm/translationTableForm.css
```diff
@@ -0,0 +1,3 @@
+.h5-translationForm {
+  font-size: 20px;  
+}
```
lencodes, line 3: I would suggest the following: because the component we're in is called `TranslationTableForm` I would prefix the `css` class with `TranslationTableForm-`. So the `h5` would become `.TranslationTableForm-h5`. This way (if implemented consistent) you can be sure to never write css rules that are targeting unwanted elements.

Missing a new line at the end of this file.

pwa/src/components/translations/translationTable.tsx
```diff
@@ -62,10 +52,10 @@ export default function TranslationTable({ tableName }) {
 
   const handleDeleteTranslation = (id): void => {
```
lencodes, line 53: This is never used. It should be. 

Maybe you can add a `Delete` button (like in the other forms tables)?

pwa/src/pages/translation-tables/[table]/translations.tsx
```diff
@@ -0,0 +1,26 @@
...
+import TranslationTable from "../../../components/translations/translationTable";
+import {HeaderContext} from "../../../context/headerContext";
+
+const IndexPage = (props) => {
+  const tableName: string = props.params.table;
+  const [__, setHeader] = React.useContext(HeaderContext);
+  
+  React.useEffect(() => {
+    setHeader({title: 'Translations', subText: 'An overview of your translations'});
+  });
```
Comment (lencodes) [inline pwa/src/pages/translation-tables/[table]/translations.tsx:11]: Max-depth errors, see other comment.

pwa/src/pages/translation-tables/[table]/translations/[id].tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const IndexPage = (props) => {
+  const id: string = props.params.id === "new" ? null : props.params.id;
+  const subtext: string = props.params.id === "new" ? 'Create a new translation' : 'View or edit this translation'
+  const tableName = props.params.table;
+  const [__, setHeader] = React.useContext(HeaderContext);
+
+  React.useEffect(() => {
+    setHeader({title: 'Translation', subText: 'View or edit this translation'});
+  });
```
Comment (lencodes) [inline pwa/src/pages/translation-tables/[table]/translations/[id].tsx:13]: Max-depth errors, see other comment.

pwa/src/pages/translation-tables/index.tsx
```diff
@@ -1,18 +1,26 @@
...
 import TableNamesTable from "../../components/translations/tableNamesTable";
+import {HeaderContext} from "../../context/headerContext";
 
 const IndexPage = () => {
+
+  const [__, setHeader] = React.useContext(HeaderContext);
+  
+  React.useEffect(() => {
+    setHeader({title: 'Translation tables', subText: 'An overview of your translation tables'});
+  });
```
lencodes, line 11: Max-depth errors, see other comment.

pwa/src/pages/translation-tables/new.tsx
```diff
@@ -0,0 +1,26 @@
...
+import TranslationTableForm from "../../components/translations/TranslationTableForm/TranslationTableForm";
+import { HeaderContext } from "../../context/headerContext";
+
+const IndexPage = () => {
+
+  const [__, setHeader] = React.useContext(HeaderContext);
+  
+  React.useEffect(() => {
+    setHeader({title: 'Translation table', subText: 'Create a new translation table'});
+  });
```
lencodes, line 11: Max-depth errors, see other comment.

lencodes (changes requested):
Looks good, left some feedback in the comments.

One other note: we're using the label "Create new" in the create buttons in these forms and tables. Can you refactor this to just "Create" to be more consistent with the rest of the application?

lencodes (approved):

PR #124, opened 2022-02-09
Configurations error 404

hghanbari:


pwa/src/apiService/apiService.ts
```diff
@@ -16,7 +16,7 @@ export default class APIService {
 
   public get adminClient (): AxiosInstance {
     return axios.create({
-      baseURL: window.GATSBY_ADMIN_URL,
+      baseURL: process.env.GATSBY_ADMIN_URL,
```
lencodes, line 19: Nice. I'm not sure who changed it to `window` but this is how it should be. Nice catch.

pwa/src/components/common/menu.tsx
```diff
@@ -59,7 +59,7 @@ export default function MainMenu() {
                     </Link>
                   </li>
                   <li className="utrecht-topnav__item">
-                    <Link to={"/configurations"} className="utrecht-topnav__link">
+                    <Link to={"/configurationsExportButton"} className="utrecht-topnav__link">
```
lencodes, line 62: This should link to `/configurations`.

pwa/src/components/configurationsExportButton/ConfigurationsExportButton.tsx
```diff
@@ -27,19 +27,23 @@ export default function ConfigurationsExportButton() {
         });
       })
       .catch((error) => {
-        throw new Error (error)
+        throw new Error(error)
```
lencodes, line 30: This should say what kind of error it is, like we do in other files.

E.g.: `throw new Error('GET export: ' + err)`

pwa/src/components/configurationsExportButton/ConfigurationsExportButton.tsx
```diff
@@ -27,19 +27,23 @@ export default function ConfigurationsExportButton() {
...
   return (
-    <button
-      className="utrecht-button"
-      type="button"
-      onClick={handleExport}
-      disabled={loading}
-    >
-      {loading ? "Perparing your download...": "Export Configuration"}
-    </button>
+    <div>
```
lencodes, line 38: This `<div>` wrapper should be removed. 

I understand why you've added it. But this layout logic should be added where the `ConfigurationsExportButton` is implemented (in this case in `/pages/configurations/index.tsx`).

pwa/src/pages/configurationsExportButton/index.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from "react"
+import Layout from "../../components/common/layout";
+import ConfigurationsExportButton from "../../components/configurationsExportButton/ConfigurationsExportButton";
+
+const IndexPage = () => {
+  return (
+    <Layout title={"ConfigurationsExportButton"} subtext={"An overview of your endpoints objects"}>
+      <title>Gateway - ConfigurationsExportButton</title>
```
lencodes, line 8: This page is just `Configurations` and not `ConfigurationsExportButton` therefore, the title should be changed.

pwa/src/pages/configurationsExportButton/index.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from "react"
```
lencodes, line 1: Adding this page was exactly what we wanted. However, on thing should be updated.

You currently named the page itself `configurationsExportButton`, but that's not a correct name for a page, because that's just the component it is (currently) including. 

The correct naming should be (and the folder names should reflect that): `configurations`

So the path of this file would become `/src/pages/configurations/index.tsx`

pwa/src/pages/configurationsExportButton/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+import Layout from "../../components/common/layout";
+import ConfigurationsExportButton from "../../components/configurationsExportButton/ConfigurationsExportButton";
+
+const IndexPage = () => {
+  return (
+    <Layout title={"ConfigurationsExportButton"} subtext={"An overview of your endpoints objects"}>
+      <title>Gateway - ConfigurationsExportButton</title>
+      <main>
+        <div className="row">
+          <div className="col-12">
```
lencodes, line 11: Note: you wrapped the `ExportConfigurationsButton` component in an extra `<div>` (presumably because the button was too wide otherwise). You should fix that in this file, instead of inside the component.

lencodes (changes requested):

pwa/src/components/configurationsExportButton/ConfigurationsExportButton.tsx
hghanbari: If i <div> remove . The button is not in the right place

PR #128, opened 2022-02-11
help button toetevoeg in alle form

hghanbari:


pwa/src/components/applications/applicationForm.tsx
```diff
@@ -130,6 +130,14 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({id}) => {
           title={title}
           cardHeader={function () {
             return (<>
+              <button
```
lencodes, line 133: **This feedback goes for every file where you've added the _Help_ button.**

@MWest2020 created a new _Help_ modal, including dynamic loading of the documentation from ReadTheDocs. We want to use this new modal within all these forms. To include this we have to do a couple of things (in each file).

**1. Create a `[document, setDocument]` state to store the documentation after the API call.**

```
const [documentation, setDocumentation] = React.useState<string>(null)
```

**2. GET the documentation using the API Service**

```
const handleSetDocumentation = (): void => {
    API.Documentation.get()
      .then((res) => {
        setDocumentation(res.data.content);
      })
      .catch((err) => {
        throw new Error("GET Documentation error: " + err);
      });
  };
```

We of course need to actually run this function at some point. You can do this within a `React.useEffect()`, preferably within an existing one. 

E.g. in `ApplicationForm` this would look like:

```
React.useEffect(() => {
    id && handleSetApplications()
    handleSetDocumentation() // we added this
  }, [API, id])
```

**3. Lastly, we need to include the actual modal**

At the same location as the previous _Help_ button, we can now add the Modal, like so:

```
<button
                className="utrecht-link button-no-style"
                data-bs-toggle="modal"
                data-bs-target="#helpModal"
                onClick={(e) => e.preventDefault()}
              >
                <Modal
                  title="Application Documentation"
                  id="helpModal"
                  body={() => (
                    <div dangerouslySetInnerHTML={{ __html: documentation }} />
                  )}
                />
```

Note: you (probably) still need to include the `Modal` component!

lencodes (changes requested):
See the inline comment 👍🏼 

If something is unclear, call me.

pwa/src/components/entities/entityForm.tsx
```diff
@@ -131,6 +142,22 @@ export const EntityForm: React.FC<EntityFormProps> = ({entityId}) => {
...
             return (
               <div>
+                <button
+                  className="utrecht-link button-no-style"
+                  data-bs-toggle="modal"
+                  data-bs-target="#entityHelpModal"
+                  onClick={(e) => e.preventDefault()}
+                >
+                  <Modal
+                    title="Entity Documentation"
```
lencodes, line 152: This should be "Object Type Documentation"

pwa/src/components/entities/entitiesTable.tsx
```diff
@@ -53,11 +53,11 @@ export default function EntitiesTable() {
             <button
               className="utrecht-link button-no-style"
               data-bs-toggle="modal"
-              data-bs-target="#helpModal"
+              data-bs-target="#entityHelpModal"
             >
               <Modal
                 title="Objects Documentation"
```
lencodes, line 59: This should be "Object Type Documentation"

pwa/src/components/objectEntities/ObjectEntityForm.tsx
```diff
@@ -134,6 +154,22 @@ export const ObjectEntityForm: React.FC<ObjectEntityFormProps> = ({objectEntityI
...
             return (
               <>
+                <button
+                  className="utrecht-link button-no-style"
+                  data-bs-toggle="modal"
+                  data-bs-target="#ObjectEntityHelpModal"
+                  onClick={(e) => e.preventDefault()}
+                >
+                  <Modal
+                    title="Entity_object Documentation"
```
lencodes, line 164: This should be "Object Documentation"

pwa/src/components/objectEntities/ObjectEntitiesTable.tsx
```diff
@@ -60,11 +60,11 @@ const ObjectEntitiesTable: React.FC<ObjectEntitiesTableProps> = ({
               <button
                 className="utrecht-link button-no-style"
                 data-bs-toggle="modal"
-                data-bs-target="#helpModal"
+                data-bs-target="#ObjectEntityHelpModal"
               >
                 <Modal
                   title="Object Entities Documentation"
```
lencodes, line 66: This should be "Object Documentation"

lencodes (approved):
Very nice @hghanbari.

See the inline comments for a couple of name changes. When updated, you can merge it 😄

PR #130, opened 2022-02-14
Handler formulier conditions refactor

hghanbari:


lencodes (changes requested):
The made changes are fine. However, they are not complete.

1. We're not sending the data to the gateway (you've created a TextArea of which we never extract the value when saving the `Handler`);

2. When we do start extracting the data and send it to the gateway, the gateway doesn't save it this is because the gateway expects an array of items (e.g. `['condition', 'condition']`); when we push an array, the gateway does save it.

Note on (2.): I think this is an issue on the backend, it shouldn't be accepting nor returning an array. However, you should double check this with Ruben. Also, if the backend is supposed to save and return arrays, then on the frontend it should not be a TextArea. 

Call me if you have any questions.

bbrands02:
Conditions is now a textfield. It gets prefilled with array/json converted to string. Then when posting the input gets parsed back to json. Also an bugfix for the description field is included otherwise this form would not work. The API calls still do fetching the old way but that is a different issue/ticket.

lencodes (commented):
Waiting on backend changes.

bbrands02:
image
Backend change has been made and is ready in the frontend

pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -173,13 +178,18 @@ export const HandlerForm: React.FC<HandlerFormProps> = ({ id, endpointId }) => {
...
+
     let url = `${context.adminUrl}/handlers`;
     let method = "POST";
     if (id) {
       url = `${url}/${id}`;
       method = "PUT";
     }
 
+
+    console.log(body);
```
lencodes: Added some unnecessary whitespace and left a `console.log` in.

pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -337,49 +347,52 @@ export const HandlerForm: React.FC<HandlerFormProps> = ({ id, endpointId }) => {
...
                       </div>
                     </div>
+                    <div className="row">
+                      <div className="col-6">
+                        <TextareaGroup
+                          name={"conditions"}
+                          id={"conditionsInput"}
+                          defaultValue={handler?.conditions}
+                          required={true}
+                        />
```
lencodes: There are a couple of issues here.

1. The gateway expects a valid JSON to be sent within the `conditions` textarea, this should reflect in the frontend.
2. The frontend is showing success alerts while getting a `422` response from the gateway (this should be showing the error alert and message).

lencodes (changes requested):
-

pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -296,49 +315,63 @@ export const HandlerForm: React.FC<HandlerFormProps> = ({ handlerId, endpointId
...
+                            return (
+                              tableNames ? (
+                                <MultiSelect
+                                  id="translationsOut"
+                                  label="Translations Out"
+                                  data={handler?.translationsOut}
+                                  options={tableNames}
+                                />
+                              ) : (
+                                <><Spinner /></>
```
lencodes: This doesn't need to be wrapped in `<></>`. It can just be the `<Spinner />` I think.

pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -296,49 +315,63 @@ export const HandlerForm: React.FC<HandlerFormProps> = ({ handlerId, endpointId
...
+                            return (
+                              tableNames ? (
+                                <MultiSelect
+                                  id="translationsIn"
+                                  label="Translations In"
+                                  data={handler?.translationsIn}
+                                  options={tableNames}
+                                />
+                              ) : (
+                                <><Spinner /></>
```
lencodes: This doesn't need to be wrapped in `<></>`. It can just be the `<Spinner />` I think.

pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -296,49 +315,63 @@ export const HandlerForm: React.FC<HandlerFormProps> = ({ handlerId, endpointId
...
                       </div>
                     </div>
+                    <div className="row">
+                      <div className="col-6">
+                        <TextareaGroup
+                          name={"conditions"}
+                          label={"Conditions (JSON)"}
+                          id={"conditionsInput"}
+                          defaultValue={handler?.conditions}
+                          required={true}
```
lencodes: When passing a boolean, you do not need to explicitly set it when it's true.

So:

`required` => defaults to `required={true}`

So only when it's false you have to actually write it out (like `required={false}`)

pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -296,49 +315,63 @@ export const HandlerForm: React.FC<HandlerFormProps> = ({ handlerId, endpointId
...
+                          infoTooltip={{
+                            content: <p>
+                              This field expects a<br/>
+                              <a
+                                target="_blank" href="https://jsonlint.com/"
+                              >
+                                valid JSON object
+                              </a>
+                            </p>
```
lencodes: The pop-up unfortunately seems to go away when I hover it. Resulting in not being able to click the actual link. I suggest removing it altogether for now.

Note: I feel as if we can remove it because you've built a well-working JSON check and error message.

pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -105,12 +115,13 @@ export const HandlerForm: React.FC<HandlerFormProps> = ({ handlerId, endpointId
     // get the inputs and check if set other set null
     let body: {} = {
```
lencodes: Because body is currently of type `{}` it does not have (e.g.) `.conditions` which is giving a Typescript error on line `143` so body should be of type `any`.

lencodes: Edit: solved.

pwa/src/components/handlers/handlerForm.tsx
lencodes: Edit: solved.

lencodes (changes requested):

PR #136, opened 2022-02-17
changed the structure of the resources

remko48:
In the file I made a small change to saveEndpoint. 
I changed the line "type: event.target.type.value," to type: "gateway-endpoint" and removed "body["type"]," in line 86. 
I did this to make sure the endpoints could still save because the gateway expects a type but we removed the type textfield.


pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -24,25 +24,27 @@ import LoadingOverlay from "../loadingOverlay/loadingOverlay";
 import APIContext from "../../apiService/apiContext";
```
lencodes, line 24: We're missing the following `const` declarations for this file to work:

```
const API: APIService = React.useContext(APIContext)
const [documentation, setDocumentation] = React.useState<string>(null)
```

pwa/src/components/objectEntities/ObjectEntityForm.tsx
```diff
@@ -22,30 +22,28 @@ import FlashMessage from 'react-flash-message';
 import LoadingOverlay from "../loadingOverlay/loadingOverlay";
```
lencodes, line 22: We're missing the `documentation` `const` declaration in this file:

```
```

Also, the `handleSetDocumentation()` is never called so it doesn't run. Could be added underneath `handleSetApplications()` on line `41`.

pwa/src/pages/endpoints/[endpointId]/handlers/[handlerId].tsx
```diff
@@ -0,0 +1,22 @@
...
+          <div className="col-12">
+            <HandlerForm id={handlerId} endpointId={endpointId} />
+          </div>
+        </div>
+      </main>
+    </Layout>
+  );
+};
+
+export default IndexPage;
```
Comment (lencodes) [inline pwa/src/pages/endpoints/[endpointId]/handlers/[handlerId].tsx:22]: Missing a white line at the end of this file (just add one enter after line `22`).

This has to do with some magic Github does for our code management (see screenshot, Github actually asks us to add it):

image

lencodes (changes requested):
Almost there! The core works great, however, some of the files are missing some imports and such and are therefore not working.

See inline comments.

Maybe we can pair-program these comments and fix the merge conflicts together 👍🏼

PR #148, opened 2022-02-22
Attribut refactor

hghanbari:


pwa/src/data/mimeTypes.ts
```diff
@@ -0,0 +1,84 @@
...
+  {name: 'application/xhtml+xml', value: 'application/xhtml+xml'},
+  {name: 'application/vnd.ms-excel', value: 'application/vnd.ms-excel'},
+  {name: 'application/vnd', value: 'application/vnd'},
+  {name: 'application/xml', value: 'application/xml'},
+  {name: 'application/vnd.mozilla.xul+xml', value: 'application/vnd.mozilla.xul+xml'},
+  {name: 'application/zip', value: 'application/zip'},
+  {name: 'video/3gpp; audio/3gpp', value: 'video/3gpp; audio/3gpp'},
+  {name: 'video/3gpp2; audio/3gpp2', value: 'video/3gpp2; audio/3gpp2'},
+  {name: 'application/x-7z-compressed', value: 'application/x-7z-compressed'},
+]
```
lencodes, line 81: I think we should rename all the `name` props of this object. We discussed earlier that we would do it the way you've built it now. However, I now realise that it's really hard to filter as a user (it's not alphabetically ordered right now).

However, the [link we used](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) is alphabetically ordered on the .extension. 

Therefore, I think we should rename all the `name` props to the `.extension` value. Like so:


```
{ name: ".aac", value: "audio/aac" },
{ name: ".abw", value: "application/x-abiword" },
```

lencodes (changes requested):
Very nice @hghanbari!!

PR #149, opened 2022-02-22
Attributes tabel refactor

hghanbari:


pwa/src/components/attributes/attributeTable.tsx
```diff
@@ -5,11 +5,12 @@ import {
   Card,
   Modal
 } from "@conductionnl/nl-design-system/lib";
-import { Link } from "gatsby";
+import {Link} from "gatsby";
```
lencodes, line 8: See how nothing actually changed here, but Github is showing the change. This is because you did not yet use `prettier`. Can you please go through all these files and save it again with `prettier` active?

pwa/src/components/attributes/attributeTable.tsx
```diff
@@ -5,11 +5,12 @@ import {
...
 import APIService from "../../apiService/apiService";
 import APIContext from "../../apiService/apiContext";
 
-export default function AttributeTable({ entityId }) {
+
```
lencodes, line 12: This whitespace shouldn't be here.

pwa/src/components/attributes/attributeTable.tsx
```diff
@@ -26,6 +27,7 @@ export default function AttributeTable({ entityId }) {
     API.Attribute.getAllFromEntity(entityId)
       .then((res) => {
         setAttributes(res.data);
+        console.log(res.data)
```
lencodes, line 30: There's still a `console.log()` here.

pwa/src/styles/main.css
```diff
@@ -236,3 +236,24 @@ a:hover {
...
+.circle.false {
+    color: #818182;
+    background-color: #e6e6e6;
+}
+
+.circle.true {
+    color: #004085;
+    background-color: #cce5ff;
+
+}
```
lencodes, line 259: We're currently using this type of label in three places: `AttributeTable`, `DashboardLogsTable` and `DashboardCard`. For each of these implementations, we're building the label from scratch (meaning we're duplicating the `html`, `ts` and `css` code three times).

I suggest making one component named (e.g.) `LabelWithBackground`, we can style this component once. When we have this component, we can then include it in `AttributeTable`, `DashboardLogsTable` and `DashboardCard`.

We could style the specific implementations with a `type` prop on the component, something like:

```
interface LabelWithBackgroundProps {
  label: string;
  type: "success" | "danger" | "primary" | "secondary";
}

const LabelWithBackground: React.FC<LabelWithBackgroundProps> = () => {
  return <></>; // create the component here
};
```

This way, we could implement the component as such:

```
<LabelWithBackground label="optional" type="secondary" />
```

and we therefore do not actually need any of the styling logic **within the implementation component** but we can keep all this logic separated.

lencodes (changes requested):
Looks very good! 🚀 

I do see some opportunities to make it a bit more technically robust and easier to maintain.

PR #154, opened 2022-02-22
GW-24 header bug

smisidjan:
Only setting the setHeader on the overview page

### DIFF
--- a/pwa/src/components/entities/entityForm.tsx
+++ b/pwa/src/components/entities/entityForm.tsx
@@ -32,14 +32,18 @@ export const EntityForm: React.FC<EntityFormProps> = ({ entityId }) => {
   const [_, setAlert] = React.useContext(AlertContext);
   const [__, setHeader] = React.useContext(HeaderContext);
 
-  React.useEffect(() => {
-    handleSetSources();
-    handleSetDocumentation();
-    entityId && handleSetEntity();
+  const handleSetHeader = () => {
     setHeader({
       title: "Object type",
       subText: "Manage your object type here"
     });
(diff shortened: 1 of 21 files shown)


pwa/src/components/entities/entityForm.tsx
```diff
@@ -32,14 +32,18 @@ export const EntityForm: React.FC<EntityFormProps> = ({ entityId }) => {
...
       title: "Object type",
       subText: "Manage your object type here"
     });
+  }
+
+  React.useEffect(() => {
+    handleSetSources();
+    handleSetDocumentation();
+    entityId && handleSetEntity();
+    handleSetHeader()
```
lencodes, line 46: Three small things to think about here:

1. Because it's **required** to run all React Hooks before any other code we usually put the hooks code above any other code. So in this case, the `handleSetHeader` method should (only in best practices, because it doesn't actually _run_ first right now, so nothing breaks) be underneath the `useEffect` hooks.

2. Because you're not actually adding any extra functionality to the `setHeader` function within the `handleSetHeader` function, I don't feel as if it should be wrapped into another function.

3. Because the header serves static information and is not dependent on (1) the `API` or (2) the `entityId` it shouldn't be based in a `useEffect` with those dependencies. You can fix this by adding a separate `useEffect` hook, like so:

```
React.useEffect(() => {
    setHeader({
      title: "Object type",
      subText: "Manage your object type here",
    });
  });
```

lencodes (changes requested):
Nice 👍🏼 

Left one small comment.

PR #155, opened 2022-02-22
GW-21 Add delete button to tables

smisidjan:
* added delete function
* added delete button to tables
* merge conflict fixes


pwa/src/components/applications/applicationsTable.tsx
```diff
@@ -46,11 +48,25 @@ export default function ApplicationsTable() {
...
       })
       .catch((err) => {
-        setAlert({message: err, type: 'danger'})
+        setAlert({ message: err, type: "danger" });
         throw new Error("GET Documentation error: " + err);
       });
   };
 
+  const handleDeleteApplication = (id): void => {
+    if (confirm(`Do you want to delete this application? With id ${id}`)) {
```
lencodes, line 57: Goes for all `confirm()` messages. I feel as if the addition of `id` in this message is unnecessary.

Firstly because the `id` is something used as an identifier for our logic, not something an end-user should be concerned about. Secondly, in most cases we cannot actually see the `id` we're referring to (we're deleting from a table in which we do not see the `id` so it's not any type of verification that the right one was deleted.

I would opt for removing the _With id ..._  part altogether.

pwa/src/pages/endpoints/[endpointId].tsx
```diff
@@ -26,15 +26,7 @@ const IndexPage = (props) => {
...
-            ) : (
-              <Tabs
-                items={[{ name: "Overview", id: "overview", active: true },
-                  {
-                    name: "Logs",
-                    id: "logs"
-                  }]}
-              />
-            )}
+            )
```
Comment (lencodes) [inline pwa/src/pages/endpoints/[endpointId].tsx:29]: Something's going wrong here. The title shows as "Logs" which should be "Endpoint". Also, there's some code showing in the frontend (probably some issues that were already there, but maybe you could fix it in this PR 👍🏼 ).

Edit: and the Log table shouldn't show underneath the form (it should only show when the tab is active).

image

lencodes (commented):
Very nice.

The `Translations` didn't receive any delete buttons.

lencodes (changes requested):
.

PR #156, opened 2022-02-23
BUGFIX ADD isLoggedIn to useEffect dependency array to mitigate max-d…

lencodes:
…epth errors


pwa/src/pages/index.tsx
```diff
@@ -7,15 +7,20 @@ const IndexPage = () => {
...
-    isLoggedIn() && (
-      <Dashboard />
-    )
-  );
+    !isLoggedIn() &&
+      setHeader({
+        subText: `Welcome to the gateway admin dashboard`,
+        title: "Dashboard",
+      });
+  }, [isLoggedIn()]);
```
lencodes, line 21: This is the only change, the rest is formatting through the new prettier.

smisidjan (approved):

PR #158, opened 2022-02-23
Objects Table columns refactor

hghanbari:


pwa/src/components/objectEntities/ObjectEntitiesTable.tsx
```diff
@@ -101,13 +101,21 @@ const ObjectEntitiesTable: React.FC<ObjectEntitiesTableProps> = ({
...
+                      headerName: "Id",
+                      field: "id"
                     },
                     {
                       headerName: "Owner",
                       field: "owner"
                     },
+                    {
+                      headerName: "Created",
+                      field: "dateCreated"
```
lencodes, line 113: We're missing the date formatting here. Currently, you're just sending the data received from the gateway to the frontend.

So, currently it's rendering: 
`2022-02-22T07:37:13+00:00`

This should be rendering as:
`2022-02-22 07:37:13`

lencodes (changes requested):

PR #159, opened 2022-02-23
Endpoint.type removed

bbrands02:


pwa/src/components/endpoints/endpointForm.tsx
```diff
@@ -190,7 +189,7 @@ export const EndpointForm: React.FC<EndpointFormProps> = ({ endpointId }) => {
                       </div>
                       <div className="col-6">
                         {
-                          applications !== null && applications.length > 0 ? (
+                          applications ? (
```
lencodes, line 192: Since we're making some changes here anyway, we might as well refactor it a bit more.

**Current code:** (with some extra comments)
```
 applications ? (
  <>
    {endpoint !== null && // could be done with an optional (?) check
    endpoint.application !== undefined && // could be done with an optional (?) check
    endpoint.application !== null ? ( // could be done with an optional (?) check
    // also: !== undefined and !== null are both falsy which means 'endpoint.application' would do the exact same 
      <SelectInputComponent
        options={applications}
        data={endpoint.application.name}
        name={"application"}
        id={"applicationInput"}
        nameOverride={"Applications"}
        value={"/admin/applications/"}
      />
    ) : (
      <SelectInputComponent
        options={applications}
        name={"application"}
        id={"applicationInput"}
        nameOverride={"Applications"}
        value={"/admin/applications/"}
      />
    )}
  </>
) : // the !applications part stays the same
```

**This code could be re-written to the following:**
_We're just passing data when we have it by doing an optional `?` check._

```
{applications ? (
  <SelectInputComponent
    options={applications}
    data={endpoint?.application?.name}
    name={"application"}
    id={"applicationInput"}
    nameOverride={"Applications"}
    value={"/admin/applications/"}
  />
) : ( // the !application part stays the same
```

Do test it before merging, but should work just fine 👍🏼

lencodes: Extra note on the comment about [Falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy?retiredLocale=nl):

We're checking the variable `endpoint.application` a couple of times:

```
if (endpoint.application !== undefined)
if (endpoint.application !== null)
```

Both checks are ensuring that `endpoint.application` is [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy). However, a truthy check can be much more simplified, like so:

```
if (endpoints.application)
```

In this case we don't even really care if it is truthy or falsy, we just want to pass the `data` prop when it exists. So we can just use the optional operator `?`, like so: `data={endpoints?.application?.name}`.

lencodes: Doesn't really matter for this pull request, but it's good to understand how this works 😄

lencodes (changes requested):

PR #160, opened 2022-02-23
Breacrumb navigation

remko48:


pwa/src/components/common/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+    setMappedCrumbs(_mappedCrumbs);
+  }, [crumbs]);
+
+  return (
+    <ul className="breadcrumbs">
+      {mappedCrumbs?.map((crumb, idx) => {
+        return (
+          <li key={idx}>
+            <Link to={crumb.pathname}>{crumb.crumbLabel.charAt(0).toUpperCase() + crumb.crumbLabel.slice(1)}</Link>
```
lencodes, line 36: Because we have the `lodash` package available in the application, we can simplify this line.

```
import * as _ from "lodash";
...
_.upperFirst(crumb.crumbLabel)
```

Instead of having to `.toUpperCase()` and `slice()` everything manually 👍🏼

lencodes (changes requested):
Left a few small comments to update and:

`ObjectEntitiesTable` and `HandlerTable` are missing `FontAwesome` imports, and are breaking because of it!

lencodes (approved):

PR #162, opened 2022-02-24
Feature/GW-405/attribute form refactor

lencodes:
Retrieves all available `attributes` instead of only the ones associated to the `Object Type` when creating and editing an `attribute` to populate the `Inversed By` select field.


remko48 (approved):

bbrands02 (commented):
image
image
Looks like inversedBy isn't being saved properly yet. Not part of this issue but could be resolved now or a bug should be created.

bbrands02:
In the dropdown it would be nice to see {entity}.{attribute} instead of only the attribute
image
Instead of Friday -> Week.Friday so that you know what entity you are joining
This isn't part of the ticket but would be a nice follow up issue

bbrands02 (approved):

lencodes:
> 

Fixed.

lencodes:
> image image Looks like inversedBy isn't being saved properly yet. Not part of this issue but could be resolved now or a bug should be created.

Fixed.

lencodes:
> In the dropdown it would be nice to see {entity}.{attribute} instead of only the attribute image Instead of Friday -> Week.Friday so that you know what entity you are joining This isn't part of the ticket but would be a nice follow up issue

The gateway doesn't seem to return this data, meaning we'd have to do many API calls to make this happen. I do agree that it'd be a nice feature. Ticket created.

PR #164, opened 2022-02-24
Feature/gw 340/rename the application to conductor

remko48:


pwa/src/pages/index.tsx
```diff
@@ -9,13 +9,13 @@ const IndexPage = () => {
...
       setHeader({
-        subText: `Welcome ${getUser().username}, to the gateway admin dashboard`,
+        subText: `Welcome ${getUser().username}, to the Conductor admin dashboard`,
         title: "Dashboard",
       });
 
     !isLoggedIn() &&
       setHeader({
-        subText: `Welcome to the gateway admin dashboard`,
+        subText: `Welcome to the Conductor admin dashboard`,
```
lencodes, line 18: Hey @remko48, this seems fine. However, could you maybe check with Ruben if this is really how we want it? I can imagine it should just be "Welcome to Conductor". Instead of the current way.

After checking (and potentially updating, depending on outcome) you can just merge this!

lencodes (changes requested):

pwa/src/pages/index.tsx
remko48: Ruben said he wants the text to be "Welcome to Conductor" so I made the changes and I will merge it.

PR #165, opened 2022-02-24
Feature/gw 60/modal overlay

MWest2020:
LINKS TO DOCS WORKING 

URL to the right documentation and path variables in files implemented.


MWest2020:
Merge conflicts resolved

lencodes (changes requested):
The parts I could review look great (very nice feature 👍🏼 ), however:

Already left some comments, however, most pages are breaking and therefore doing a review is difficult. 

I see that `API.Documentation.get()` is called **without** passing the `path` prop which breaks the code (resulting in a `404`) in at least the following files: 

- handlerForm;
- handlerTable;
- logTable and
- usersTable.

Also, in more files it's being called with an empty `path` prop (`""`). I'm unsure if that's intended.

In some files the `handleSetDocumentation()` method is called multiple times (at least in `attributeTable`).

The edit and create source routes break (I'm assuming because of a broken `Documentation.get()` call).

Please request a re-review after fixing these issues.

PR #166, opened 2022-02-25
GW-260 fix error object entities

smisidjan:
* added validate uuid package
* added valid uuid check on external id 
* fix select application


bbrands02:
This form me be replaced by form.io but the changes are valid

bbrands02 (approved):

PR #172, opened 2022-02-28
Handler implement with new API Service

hghanbari:
Document is niet gewerkt . @mark gaat gedaan. Dit PR moet wachten tot met PR #130 is gemerg worden.


pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -26,104 +25,72 @@ import MultiSelect from "../common/multiSelect";
...
   const [tableNames, setTableNames] = React.useState<Array<any>>(null);
-  const title: string = id ? "Edit Handler" : "Create Handler";
+  const title: string = handlerId ? "Edit Handler" : "Create Handler";
   const API: APIService = React.useContext(APIContext);
   const [documentation, setDocumentation] = React.useState<string>(null);
   const [_, setAlert] = React.useContext(AlertContext);
   const [__, setHeader] = React.useContext(HeaderContext);
 
   React.useEffect(() => {
+    // handleSetDocumentation();
```
lencodes: When blocks of code aren't working, we might as well just delete them. 

Therefore, I suggest deleting this line of code and it's corresponding functions (starting on line `96`) and state (on line `40`).

pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -26,104 +25,72 @@ import MultiSelect from "../common/multiSelect";
 import ElementCreationNew from "../common/elementCreationNew";
 
 interface HandlerFormProps {
-  id: string;
+  handlerId: string;
```
lencodes, line 28: You've changed this prop key, but you have not changed the key when passing it to this component.

Meaning within `[handlerId].tsx`: 

Current: `<HandlerForm id={handlerId} endpointId={endpointId} />`

Target: `<HandlerForm {... {handlerId, endpointId}} />`

lencodes (changes requested):

PR #175, opened 2022-03-01
Feature/gw 343/prettify

bbrands02:


pwa/src/styles/prism.css
```diff
@@ -0,0 +1,3 @@
...
...ze:1em;text-align:left;white-space:pre;word-spacing:normal;word-break:normal;word-wrap:normal;line-height:1.5;-moz-tab-size:4;-o-tab-size:4;tab-size:4;-webkit-hyphens:none;-moz-hyphens:none;-ms-hyphens:none;hyphens:none}pre[class*=language-]{padding:1em;margin:.5em 0;overflow:auto;border-radius:.3em}:not(pre)>code[class*=language-],pre[class*=language-]{background:#272822}:not(pre)>code[class*=language-]{padding:.1em;border-radius:.3em;white-space:normal}.token.cdata,.token.comment,.token.doctype,.token.prolog{color:#8292a2}.token.punctuation{color:#f8f8f2}.token.namespace{opacity:.7}.token.constant,.token.deleted,.token.property,.token.symbol,.token.tag{color:#f92672}.token.boolean,.token.number{color:#ae81ff}.token.attr-name,.token.builtin,.token.char,.token.inserted,.token.selector,.token.string{color:#a6e22e}.language-css .token.string,.style .token.string,.token.entity,.token.operator,.token.url,.token.variable{color:#f8f8f2}.token.atrule,.token.attr-value,.token.class-name,.token.function{color:#e6db74}.token.keyword{color:#66d9ef}.token.important,.token.regex{color:#fd971f}.token.bold,.token.important{font-weight:700}.token.italic{font-style:italic}.token.entity{cursor:help}
```
lencodes, line 3: Shouldn't this be included in the `prismjs` package? Why are we saving this `css` in our own codebase?

pwa/src/components/common/codeBlock/codeBlock.tsx
```diff
@@ -1,19 +1,27 @@
...
 
-  return (
-    <pre className="codeBlock">
-      <code className={`codeBlock-code--${language}`}>
+  React.useEffect(() => {
+    Prism.highlightAll();
+  }, [Prism])
+
+  return (<>
+    <pre>
```
lencodes, line 20: Since we have a `pre` element wrapping the entire component, we do not need the empty wrapper (`<></>`).

pwa/src/components/common/codeBlock/codeBlock.tsx
```diff
@@ -1,19 +1,27 @@
...
 import "./codeBlock.css";
+import Prism from "prismjs";
+import "../../../styles/prism.css";
+import 'prismjs/components/prism-json';
+import 'prismjs/components/prism-yaml';
 
 interface CodeBlockProps {
   code: any,
-  language: "json" | "xml"
+  language: "json" | "xml" | "css" | "html" | "yaml"
```
lencodes: I don't think we're actually ever rendering `css` `html` or `yaml` code. Therefore, I would not add them to the language options here.

pwa/src/components/logs/logModal/LogModal.tsx
```diff
@@ -10,7 +10,25 @@ interface LogModalProps {
...
+    log?.requestHeaders['content-type'] && setResponseCodeLanguage(getCodeLanguage(log?.requestHeaders['content-type'][0]));
+  }, [log]);
+
+  const getCodeLanguage = (header: string) => {
+    if (header === 'application/form.io') {
+      return 'json'
+    } 
+    
+    return header.replace('application/', '')
+  };
```
lencodes: 1. I think the `log` object **should never be null or undefined** within the `LogModal` component. Therefore, any and all checks if `log` exists (`.?`) shouldn't be here (this means that the implementation of `LogModal` might need updating.

2. Similarly, should a `log` always have `.requestHeaders.accept` and `.requestHeaders['content-type']`? If so, all these optional checks should be removed.

3. With (1) and (2) updated, we can just start the function (e.g. `setRequestCodeLanguage(log.requestHeaders.accept[0]`).

4. The logic of `getCodeLanguage()` might be better placed within the `/services` folder and they could get a bit more descriptive names (e.g. `getRequestAcceptHeader()` and `getRequestContentTypeHeader()`). _Note: you might have to then alter the response you get (e.g. the `if (formio) return json) within the `LogModal` itself_.

lencodes (changes requested):

pwa/src/styles/prism.css
bbrands02: This is how prismjs works because if all css would be in the prismjs package it would be enormous with a lot of css that will never be used for many supported languages

pwa/src/components/logs/logModal/LogModal.tsx
bbrands02: 2. requestHeaders could always differ because of user input, therefore the ? checks should stay

smisidjan (approved):

PR #177, opened 2022-03-01
GW-442 Added subscriber pages

smisidjan:
added new subscriber pages

the cleanup branch must be pulled first


pwa/src/apiService/apiService.ts
```diff
@@ -14,6 +14,7 @@ import Test from "./resources/test";
 import Handler from "./resources/handler";
 import ApiCalls from "./resources/apiCalls";
 import { GATSBY_ADMIN_URL, GATSBY_API_URL } from "../../static/env.js";
+import Subscriber from "./resources/subscriber";
```
lencodes, line 16: If you move this addition one line up, all resource imports will be after one another.

pwa/src/apiService/resources/subscriber.ts
```diff
@@ -0,0 +1,33 @@
...
+export default class Subscriber {
+  private _instance: AxiosInstance;
+
+  constructor(_instance: AxiosInstance) {
+    this._instance = _instance;
+  }
+
+  public getAll = (): Promise<AxiosResponse> => {
+    return this._instance.get("/subscribers");
+  };
```
lencodes: This is not implemented anywhere. Therefore, I would remove it (no unused code).

pwa/src/apiService/resources/subscriber.ts
```diff
@@ -0,0 +1,33 @@
...
+    return this._instance.put(`/subscribers/${id}`, JSON.stringify(data));
+  };
+
+  public getAllFromEntity = (entityId: string): Promise<AxiosResponse> => {
+    return this._instance.get(`/subscribers?entity.id=${entityId}`);
+  };
+
+  public delete = (id: string): Promise<AxiosResponse> => {
+    return this._instance.delete(`/subscribers/${id}`)
+  }
```
lencodes, line 28: This is not implemented anywhere (but I think it should be).

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
+import * as React from "react";
+import { Link } from "gatsby";
+import {
+  checkValues,
```
lencodes, line 4: This is not implemented anywhere. Therefore, I would remove it (no unused code).

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+  Checkbox,
+  SelectInputComponent,
+  TextareaGroup,
+  Accordion,
+  Spinner,
+  Card,
+  Modal,
+} from "@conductionnl/nl-design-system/lib";
+import { navigate } from "gatsby-link";
+import ElementCreationNew from "../common/elementCreationNew";
```
lencodes: This is not implemented anywhere. Therefore, I would remove it (no unused code).

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+import APIService from "../../apiService/apiService";
+import APIContext from "../../apiService/apiContext";
+import LoadingOverlay from "../loadingOverlay/loadingOverlay";
+import { AlertContext } from "../../context/alertContext";
+import { HeaderContext } from "../../context/headerContext";
+import { MIMETypes } from "../../data/mimeTypes";
```
lencodes: This is not implemented anywhere. Therefore, I would remove it (no unused code).

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+import MultiSelect from "../common/multiSelect";
+
+interface ISubscriber {
```
lencodes: The naming convention for the `props` interface is: `ComponentNameProps`. So in this case: `SubscriberFormProps`.

Only when we're adding other interfaces (e.g. for a data type within the component) we start using the `I...` naming convention.

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+    });
+  }, [setHeader]);
+
+
+  React.useEffect(() => {
+    subscriberId && handleSetSubscriber()
+    handleSetSources();
+    handleSetEndpoints();
+    handleSetTableNames();
+  }, [API]);
```
lencodes: I think this should re-run when `subscriberId` changes. Therefore, I would add `subscriberId` to the dependency array.

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+        setShowSpinner(false);
+      });
+  };
+
+  const handleSetTableNames = () => {
+    setShowSpinner(true);
+
+    API.Translation.getTableNames()
+      .then((res) => {
+        const convertedArray = res.data["results"].map((value, idx) => ({ id: idx, name: value, value: value }));
```
lencodes: 1. We're getting the `.results` part of this object with the `["key"]` syntax here. However, you can just use the `.results` syntax here (which would be easier to read and more consistent with the rest of the code base. This would result in `res.data.results.map(....)`.
2. The const name `convertedArray` doesn't say much about the array itself (e.g. what does it contain). I would suggest making a more descriptive name (e.g. `mappedTableNames`).

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+      <Card
+        title={title}
+        cardHeader={function () {
+          return (
+            <>
+              <Link className="utrecht-link" to={`/entities/${entityId}`}>
+                <button className="utrecht-button utrecht-button-sm btn-sm btn btn-light mr-2">
+                  <i className="fas fa-long-arrow-alt-left mr-2" />
+                  Back
+                </button>
```
lencodes, line 217: This can now link back to the `Subscribers` tab in the page. Like so:

```
<Link className="utrecht-link" to={`/entities/${entityId}`} state={{ activeTab: "subscribers" }}>
  ...
</Link>
```

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+    };
+
+    body = removeEmptyObjectValues(body);
+
+    if (!subscriberId) {
+      // unset id means we're creating a new entry
+      API.Subscriber.create(body)
+        .then(() => {
+          setAlert({ message: "Saved subscriber", type: "success" });
+          navigate(`/entities/${entityId}`);
```
lencodes: This can now redirect to the `Subscribers` tab in the page, like so:

```
navigate(`/entities/${entityId}`, {
  state: { activeTab: "subscribers" },
});
```

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+                            { name: "POST", value: "POST" },
+                            { name: "PUT", value: "PUT" },
+                            { name: "GET", value: "GET" },
+                          ]}
+                          name={"method"}
+                          id={"methodInput"}
+                          nameOverride={"Method"}
+                          data={subscriber && subscriber.method && subscriber.method}
+                          required
+                        />
```
lencodes: This select component is not showing the right data (the backend does return the right data).

Note: debug this when doing a hard refresh (the right data _is_ saved in the `state` and is therefore showing correctly after saving but not after reloading).

lencodes: Note: the other fields that don't render correctly, I'll just add a small comment but the same feedback might apply.

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+                      </div>
+                      <div className="col-6">
+                        <GenericInputComponent
+                          type={"number"}
+                          name={"order"}
+                          id={"orderInput"}
+                          data={subscriber && subscriber.order && subscriber.order}
+                          nameOverride={"Order"}
+                        />
```
lencodes: This doesn't render the saved value.

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+                    <br/>
+                    <div className="row">
+                      <div className="col-6">
+                        <GenericInputComponent
+                          type={"text"}
+                          name={"conditions"}
+                          id={"conditionsInput"}
+                          data={subscriber && subscriber.conditions && subscriber.conditions}
+                          nameOverride={"Conditions"}
+                        />
```
lencodes: 1. This field must be a valid JSON object for it to be processable in the backend. However, (1) its not mentioned anywhere so the user doesn't know and (2) when the user adds another type of value a vague error message is returned. This is resulting in a form that cannot be submitted if you do not already know that it has to be a JSON.
2. This may become a textarea instead of an regular input.

Note: @hghanbari had the same issue regarding the user feedback, maybe its best to quickly check with him how he's fixing it so that you can use that fix as well (for consistency).

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+                      </div>
+                    </div>
+                    <br/>
+                    <div className="row">
+                      <div className="col-6">
+                        <GenericInputComponent
+                          type={"text"}
+                          name={"conditions"}
+                          id={"conditionsInput"}
+                          data={subscriber && subscriber.conditions && subscriber.conditions}
```
lencodes: I think the second `&& subscriber.conditions` isn't doing anything. Even the first one is unnecessary, I think this should work:

`data={subscriber?.conditions}`

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+                          nameOverride={"Conditions"}
+                        />
+                      </div>
+                      <div className="col-6">
+                        {sources !== null && sources.length > 0 ? (
```
lencodes: This doesn't render the saved value.

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+                            ) : (
+                              <SelectInputComponent
+                                options={sources}
+                                name={"source"}
+                                id={"sourceInput"}
+                                nameOverride={"Source"}
+                                value={"/admin/sources/"}
+                              />
+                            )}
+                          </>
```
lencodes: This entire duplication of code is based on if `subscriber` exists, if it does, we're passing a `data` prop.

I think this would do the same (and we've been implementing it like this in some other forms, seems to work fine):


```
<SelectInputComponent
  options={sources}
  data={subscriber?.source.name} // this "?" makes it an optional
  name={"source"}
  id={"sourceInput"}
  nameOverride={"Source"}
  value={"/admin/sources/"}
/>
```

This optional checks if `subscriber` exists (or is [Truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy?retiredLocale=nl)), this removes the need to check `subscriber !== null`.

After that, we're also checking if `subscriber.source` is not `null` or `undefined`. I think we don't have to do this check at all (it should always return `.source`, but if it doesn't we could add the optional `?` operator once more).

So it might have to be updated to `data={subscriber?.source?.name}` (but I think we can leave the second optional operator out of it).

Finally, we don't have to do the `? <> : <>` anymore. Because when we pass the `data` prop **without** `subscriber` available, we're just passing `null` which will render it without data.

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+                            id={"sourceInput"}
+                            nameOverride={"Source"}
+                          />
+                        )}
+                      </div>
+                    </div>
+                    <br/>
+                    <div className="row">
+                      <div className="col-6">
+                        {endpoints !== null && endpoints.length > 0 ? (
```
lencodes: This doesn't render the saved value.

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+                            ) : (
+                              <SelectInputComponent
+                                options={endpoints}
+                                name={"endpoint"}
+                                id={"endpointInput"}
+                                nameOverride={"Endpoint"}
+                                value={"/admin/endpoints/"}
+                              />
+                            )}
+                          </>
```
lencodes: This should also become a `?` check instead of this.

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+                      <div className="col-12 col-sm-6 ">
+                        <div className="form-check">
+                          <Checkbox
+                            type={"checkbox"}
+                            id={"asynchronousInput"}
+                            nameLabel={"Asynchronous"}
+                            nameAttribute={"asynchronous"}
+                            data={subscriber && subscriber.asynchronous && subscriber.asynchronous}
+                            defaultValue={"true"}
+                          />
```
lencodes: This doesn't render the saved value.

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+                    <br/>
+                    <div className="row mt-3">
+                      <div className="col-12 col-sm-6 ">
+                        <div className="form-check">
+                          <Checkbox
+                            type={"checkbox"}
+                            id={"asynchronousInput"}
+                            nameLabel={"Asynchronous"}
+                            nameAttribute={"asynchronous"}
+                            data={subscriber && subscriber.asynchronous && subscriber.asynchronous}
```
lencodes: This can just be `subscriber?.asynchronous`.

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+                      <div className="col-12 col-sm-6 ">
+                        <div className="form-check">
+                          <Checkbox
+                            type={"checkbox"}
+                            id={"blockingInput"}
+                            nameLabel={"Blocking"}
+                            nameAttribute={"blocking"}
+                            data={subscriber && subscriber.blocking && subscriber.blocking}
+                            defaultValue={"true"}
+                          />
```
lencodes: This doesn't render the saved value.

Here the data prop can also be updated.

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+                                      },
+                                    ]
+                                    : null
+                                }
+                              />
+                            );
+                          },
+                        },
+                      ]}
+                    />
```
lencodes, line 468: All of these fields are breaking.

image

pwa/src/components/subscribers/subscribersTable.tsx
```diff
@@ -0,0 +1,139 @@
...
+        throw new Error("GET Subscribers error: " + err);
+      })
+      .finally(() => {
+        setShowSpinner(false);
+      });
+  };
+
+  const handleDeleteSubscriber = (id): void => {
+    if (confirm(`Do you want to delete this subscriber?`)) {
+      API.Entity.delete(id)
```
lencodes: This should be `API.Subscriber` I assume.

pwa/src/components/subscribers/subscribersTable.tsx
```diff
@@ -0,0 +1,139 @@
...
+                      headerName: "Name",
+                      field: "name",
+                    },
+                    {
+                      headerName: "Method",
+                      field: "method",
+                    },
+                    {
+                      headerName: "Endpoint",
+                      field: "endpoint",
```
lencodes, line 91: This is now rendering the actual endpoint, it might be better to show the `name` here.

(You also select the name when creating the subscriber, so it might be a better reference point).

pwa/src/components/subscribers/subscribersTable.tsx
```diff
@@ -0,0 +1,139 @@
...
+                <Table
+                  columns={[
+                    {
+                      headerName: "Name",
+                      field: "name",
+                    },
+                    {
+                      headerName: "Method",
+                      field: "method",
+                    },
```
lencodes: It might be nice to render a `LabelWithBackground` component here.

lencodes (changes requested):

pwa/src/components/subscribers/subscribersTable.tsx
smisidjan: This is not yet possible because something is still going wrong in the backend, there has to be a fix for that before I can show the name. There is already a function for that.

PR #179, opened 2022-03-02
Added Breadcrumbs to Objects

remko48:


pwa/src/components/objectEntities/ObjectEntityFormNew.tsx
```diff
@@ -9,11 +9,11 @@ import LoadingOverlay from "../loadingOverlay/loadingOverlay";
 import { Link } from 'gatsby';
 
 interface ObjectEntityFormNewProps {
-  objectEntityId: string,
+  objectId: string,
   entityId: string,
 }
 
-export const ObjectEntityFormNew: React.FC<ObjectEntityFormNewProps> = ({ objectEntityId, entityId }) => {
+export const ObjectEntityFormNew: React.FC<ObjectEntityFormNewProps> = ({ objectId, entityId }) => {
```
lencodes, line 16: Why is there a `ObjectEntityFormNew`? I see that the `ObjectEntityForm` still exists in our codebase but is never implemented.

Could we just delete the old one and call this one `ObjectEntityForm`? Instead of cluttering our codebase. 

Maybe you can (1) check  (I think with @bbrands02) if I'm correct and fix it immediately or (2) create a JIRA ticket.

pwa/src/components/objectEntities/ObjectEntityFormNew.tsx
```diff
@@ -9,11 +9,11 @@ import LoadingOverlay from "../loadingOverlay/loadingOverlay";
 import { Link } from 'gatsby';
 
 interface ObjectEntityFormNewProps {
-  objectEntityId: string,
+  objectId: string,
```
lencodes, line 12: Nice catch 👍🏼

pwa/src/pages/entities/[entityId]/objects/[objectId].tsx
```diff
@@ -37,7 +37,7 @@ const IndexPage = (props) => {
                 aria-labelledby="overview-tab"
               >
                 <br />
-                <ObjectEntityFormNew entityId={entityId} objectEntityId={objectEntityId} />
+                <ObjectEntityFormNew entityId={entityId} objectId={objectId} />
```
Comment (lencodes) [inline pwa/src/pages/entities/[entityId]/objects/[objectId].tsx]: We could spread these props, like so: `<ObjectEntityFormNew {...{entityId, objectId}} />`

lencodes (changes requested):
Looks good to me, couple of small comments.

pwa/src/components/objectEntities/ObjectEntityFormNew.tsx
remko48: I spoke with @bbrands02 and the file wil be used and renamed. That wil be done under an already existing JIRA ticket.

lencodes (approved):
Looks good to me.

PR #181, opened 2022-03-02
GW-351

smisidjan:
updated multi select component
added new function to the inputHandler


pwa/src/components/attributes/attribuutForm.tsx
```diff
@@ -664,6 +664,7 @@ export const AttributeForm: React.FC<AttributeFormProps> = ({ attributeId, entit
                           title: "File Types",
                           id: "fileTypesAccordion",
                           render: function () {
+                            console.log(attribute?.fileTypes)
```
lencodes: Missed a console.log

pwa/src/components/common/multiSelect.tsx
```diff
@@ -18,40 +18,52 @@ interface ISelectOption {
...
+  };
+
   const generateId = (): string => Math.random().toString(36);
 
   return (
     <div>
       {values.length > 0 &&
       <ul style={{ paddingLeft: 0 }}>
         {values.map((value, idx) => {
+          console.log({ values })
```
lencodes: Missed a console.log

lencodes (approved):
Nice 👍🏼 

I think, once the MVP is done, we should take another look at these changes. For now, I don't really see another easy way to fix this issue. However, I do think that this contains some logic that hard to maintain and hard to expand upon.

PR #185, opened 2022-03-02
Gw 60/handler logtable patch

MWest2020:
This should be the correct path without breaking anything else.


pwa/src/components/handlers/handlerTable.tsx
```diff
@@ -36,7 +36,7 @@ export default function HandlersTable({ endpointId }) {
   };
 
   const handleSetDocumentation = (): void => {
-    API.Documentation.get("attributes")
+    API.Documentation.get("handlers")
```
lencodes, line 39: This is not returning any content.

(Might be a known issue)

pwa/src/components/logs/logTable/logTable.tsx
```diff
@@ -25,6 +25,21 @@ export const LogTable: React.FC<LogTableProps> = ({ entityId, sourceId, endpoint
...
   }, [API, entityId, endpointId, sourceId]);
 
+  React.useEffect(() => {
+    handleSetDocumentation();
+  });
+
+  const handleSetDocumentation = (): void => {
+    API.Documentation.get("logtables")
+      .then((res) => {
+        setDocumentation(res.data.content);
```
lencodes, line 35: This `state` (`const [documentation, setDocumentation] ...`) does not exist and therefore the application is breaking.

Also, when re-implementing this state please check the modal, because it's not implementing `documentation` as content anymore either.

**Note:** this was intentionally removed because this functionality was commented out in `development` instead of removed.

lencodes (changes requested):
Left a couple of notes. I think in the merging of other branches, this one broke.

Also, I see that this ticket `GW-60` is already on review (since it's still in this phase, it can't be in that column I think). I think this might better have been a new bug ticket.

pwa/src/components/handlers/handlerTable.tsx
MWest2020: handlers doesn't have content yet

PR #187, opened 2022-03-02
added condition to check if table is empty

remko48:


pwa/src/components/translations/translationTable.tsx
```diff
@@ -38,6 +38,7 @@ export default function TranslationTable({ tableName }) {
     setShowSpinner(true);
     API.Translation.getAllFrom(tableName)
       .then((res) => {
+        {res.data.length <=  0 && navigate("/translation-tables")}
```
lencodes: I'm thinking this might not be the best place to do this check. Because right now, the following flow happens (I think):

1. We `DELETE` a translation table;
2. we `GET` all translation tables;
3. we notice that we do not have any (we get a empty `[]` I suppose) and only thereafter 
4. we `navigate` away.

Wouldn't it be better to do the following:

1. We `DELETE` a translation table;
2. we check if we just deleted the last one and then
3. we `navigate` away 

I this second scenario, we're not doing an extra call (and we don't linger on the page for another call, making the flow much smoother).

lencodes: To implement this, we can add the following to the `handleDeleteTranslation()` function (within the `.then()` callback):

```
setAlert({ message: `Deleted translation with id: ${id}`, type: "success" });

if (translations.length === 1) { // removed the last translation table <-- keep this comment I think
  navigate("/translation-tables")
}

getTranslations(); // if length wasn't 1 we still do the same flow as before <-- remove this comment
```

Another possible syntax would be: `translations.length === 1 && navigate(...)`

lencodes (changes requested):
Nice catch 👍🏼

pwa/src/components/translations/translationTable.tsx
```diff
@@ -49,12 +51,18 @@ export default function TranslationTable({ tableName }) {
...
+      setLoadingOverlay(true);
       API.Translation.delete(id)
         .then(() => {
           setAlert({ message: `Deleted translation with id: ${id}`, type: "success" });
+          translations.length === 1 && navigate("/translation-tables")// removed the last translation table
           getTranslations();
         })
+        .finally(() => {
+          setLoadingOverlay(false);
+        })
```
lencodes: I'd put this underneath the `.catch()` for better readability. The `finally()` is last to be called and therefore I would say logically underneath everything.

lencodes (approved):
No need for a re-review, you can merge it 👍🏼

PR #190, opened 2022-03-03
HEADER names added

MWest2020:
NO merge conflicts, but IDK if ok. Formatter doesn't work

### DIFF
--- a/pwa/src/components/applications/applicationForm.tsx
+++ b/pwa/src/components/applications/applicationForm.tsx
@@ -16,7 +16,6 @@ import APIContext from "../../apiService/apiContext";
 import LoadingOverlay from "../loadingOverlay/loadingOverlay";
 import { HeaderContext } from "../../context/headerContext";
 import { AlertContext } from "../../context/alertContext";
-import ApplicationsTable from "./applicationsTable";
 
 interface IApplication {
   name: string;
@@ -43,8 +42,7 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({ id }) => {
 
   React.useEffect(() => {
     setHeader({
-      title: `Applications ${application ? application.name : ""}`,
-      subText: "Manage your applications here",
+      title: `Application ${application ? application.name : ""}`
     });
   }, [setHeader, application]);
 
--- a/pwa/src/components/applications/applicationForm.tsx
+++ b/pwa/src/components/applications/applicationForm.tsx
@@ -16,6 +16,7 @@ import APIContext from "../../apiService/apiContext";
 import LoadingOverlay from "../loadingOverlay/loadingOverlay";
 import { HeaderContext } from "../../context/headerContext";
 import { AlertContext } from "../../context/alertContext";
+import ApplicationsTable from "./applicationsTable";
 
 interface IApplication {
   name: string;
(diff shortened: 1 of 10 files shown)


pwa/src/components/applications/applicationForm.tsx
```diff
@@ -16,6 +16,7 @@ import APIContext from "../../apiService/apiContext";
 import LoadingOverlay from "../loadingOverlay/loadingOverlay";
 import { HeaderContext } from "../../context/headerContext";
 import { AlertContext } from "../../context/alertContext";
+import ApplicationsTable from "./applicationsTable";
```
lencodes: I don't think we're actually using this import.

pwa/src/components/applications/applicationForm.tsx
```diff
@@ -42,10 +43,10 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({ id }) => {
 
   React.useEffect(() => {
     setHeader({
-      title: "Applications",
+      title: `Applications ${application ? application.name : ""}`,
```
lencodes: This should be singular (we're currently in a single **application** and not in applications).

pwa/src/components/attributes/attributeForm.tsx
```diff
@@ -275,7 +275,7 @@ export const AttributeForm: React.FC<AttributeFormProps> = ({ attributeId, entit
                     </div>
                     <br />
                     <div className="row">
-                      <div className="col-6">
+                      {/* <div className="col-6">
```
lencodes: This shouldn't be commented out.

pwa/src/pages/entities/[entityId]/attributes/[attributeId].tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from "react";
-import AttributeForm from "../../../../components/attributes/attribuutForm";
+import AttributeForm from "../../../../components/attributes/attributeForm";
```
Comment (lencodes) [inline pwa/src/pages/entities/[entityId]/attributes/[attributeId].tsx:2]: 👍🏼

lencodes (changes requested):
Two global things we should add to this pull request:

1. Currently, the `HeaderContext` accepts as the `title: string` a string, if we make this a `title: string | JSX.Element` we can change your `setHeading` syntax throughout all the files you've updated to (e.g.): 

```
setHeader({
  title: <>Application {application && <i>{application.name}</i>}</>,
  subText: "Manage your applications here",
});
```

This way, its italic and we can remove the `? <> : <>` syntax.

Note this would mean updating the `headerContext.ts` props to:

```
export interface HeaderProps {
  title: string | JSX.Element;
  subText: string;
}
```

2. Since we've added breadcrumbs, the `subText` in the `setHeader()` function is actually **never used**. So we can remove this by (1.) removing the `subText` prop from the `HeaderContext` and (2.) removing the `subText` prop from every `setHeader()` call (throughout all files, not just the ones changed in this pull request).

Edit: this would mean that the final implementation would be (e.g.):

```
setHeader({
  title: <>Application {application && <i>{application.name}</i>}</>,
});
```

pwa/src/components/applications/applicationForm.tsx
MWest2020: check

PR #193, opened 2022-03-03
Alle logs table vervangen met de nieuwe LogsTable

hghanbari:


pwa/src/components/logs/logModal/LogModal.tsx
```diff
@@ -38,10 +38,10 @@ const LogModal: React.FC<LogModalProps> = ({ log }) => {
...
                       <th>Status</th>
                       <td>
-                        <StatusCode
-                          code={log?.responseStatusCode}
-                          message={log?.responseStatus}
-                        />
+                        {/*<StatusCode*/}
+                        {/*  code={log?.responseStatusCode}*/}
+                        {/*  message={log?.responseStatus}*/}
+                        {/*/>*/}
```
lencodes: We're now missing the status code of the log in the modal. We should add the `LabelWithBackground` here, like so:

`<LabelWithBackground label={log?.responseStatusCode} type={statusClass} />`

You can create the `statusClass` within the `body` function on line `21`, like so:

`const statusClass = log.responseStatusCode > 199 && log.responseStatusCode < 300 ? "success" : "danger";`

pwa/src/components/logs/logTable/logTable.css
```diff
@@ -1,3 +1,41 @@
...
+    font-size: 16px;
+}
+
+.logsTable .logsTable-viewLogTd {
+    display: flex;
+    justify-content: right;
+}
 .logTable-statusCode {
     font-size: '10px' !important;
 }
```
lencodes: All `.logTable-statusCode` lines can be removed (we're not styling this statuscode here, we're getting the styling from `LabelWithBackground`).

This means you can remove:

```
.logTable-statusCode {
    font-size: '10px' !important;
}

.logTable-statusCode--red {
    color: red;
}

.logTable-statusCode--green {
    color: limegreen;
}
```

pwa/src/components/logs/logTable/logTable.tsx
```diff
@@ -1,218 +1,63 @@
...
-    <span>
-      <FontAwesomeIcon className={`logTable-statusCode--${statusColor} mr-2`} icon={faCircle} />
-      {`${code} (${message})`}
-    </span>
-  );
-};
-
-export default LogTable;
+export class StatusCode {
+}
```
lencodes: This is now redundant and can be removed.

pwa/src/components/logs/logTable/logTable.tsx
```diff
@@ -1,218 +1,63 @@
...
+        <thead>
+        <tr>
+          <th>Status</th>
+          <th>Type</th>
+          <th>Method</th>
+          <th>Response time</th>
+        </tr>
+        </thead>
+        <tbody>
+        {!logs.length && <tr><td>No results found</td></tr> }
```
lencodes: I think we should move this line to the bottom of `tbody` instead of right at the top.

Because: the core of this component is showing the logs, this only kicks in when there are no logs available, which is the edge case and therefore makes more sense to be at the bottom.

pwa/src/pages/endpoints/[endpointId].tsx
```diff
@@ -3,10 +3,28 @@ import { Tabs } from "@conductionnl/nl-design-system/lib/Tabs/src/tabs";
...
+    handleSetLogs();
+  },[API]);
+
+  const handleSetLogs = (): void => {
+    API.Log.getAllFromEndpoint(endpointId)
+      .then((res) => {
+        setLogs(res.data);
+      })
+      .catch((err) => {
+        throw new Error(`GET Logs error: ${err}`);
```
Comment (lencodes) [inline pwa/src/pages/endpoints/[endpointId].tsx]: This can specify that it's an error getting the Endpoint logs, so: `GET Endpoint Logs error: ...`

pwa/src/pages/endpoints/[endpointId].tsx
```diff
@@ -51,7 +73,7 @@ export const IndexPage = (props) => {
             </div>
             <div className="tab-pane" id="logs" role="tabpanel" aria-labelledby="logs-tab">
               <br />
-              <LogTable {...{ endpointId }} />
+              <LogTable logs={logs} />
```
Comment (lencodes) [inline pwa/src/pages/endpoints/[endpointId].tsx]: You could use the following syntax here (we've been using it a lot, and for consistency I'd recommend using it here as well):

`<LogsTable {...{logs}} />`

Note: you can do this because the name and value are both the same (in this case `logs).

pwa/src/pages/entities/[entityId].tsx
```diff
@@ -67,7 +84,7 @@ const IndexPage = (props) => {
             </div>
             <div className="tab-pane" id="logs" role="tabpanel" aria-labelledby="logs-tab">
               <br />
-              <LogTable {...{ entityId }} />
+              <LogTable logs={logs}  />
```
Comment (lencodes) [inline pwa/src/pages/entities/[entityId].tsx]: Can also become `{...{logs}}`

pwa/src/pages/entities/[entityId].tsx
```diff
@@ -5,11 +5,28 @@ import EntityForm from "../../components/entities/entityForm";
...
+    handleSetLogs();
+  },[API]);
+
+  const handleSetLogs = (): void => {
+    API.Log.getAllFromEntity(entityId)
+      .then((res) => {
+        setLogs(res.data);
+      })
+      .catch((err) => {
+        throw new Error(`GET Logs error: ${err}`);
```
Comment (lencodes) [inline pwa/src/pages/entities/[entityId].tsx]: Can also get a more descriptive error message

pwa/src/pages/endpoints/[endpointId].tsx
Comment (lencodes) [reply pwa/src/pages/endpoints/[endpointId].tsx]: Note: goes for all other files as well

pwa/src/pages/endpoints/[endpointId].tsx
Comment (lencodes) [reply pwa/src/pages/endpoints/[endpointId].tsx]: Note: goes for all other files as well

Comment (lencodes) [inline pwa/src/pages/endpoints/[endpointId].tsx]: Currently, the LogTable is rendering on a grey background and without the right layout (padding and margins and such). We can fix this, by adding a `Card` component around it (just like other tables have), like so:

```
<Card title="Endpoint Logs" cardHeader={() => <></>} cardBody={() => <LogTable logs={logs} />} />
```

You can also add a refresh button in the `cardHeader={() => ...}`, in the above example I'm just adding a empty object.

Note: this also goes for all the other files where you've implemented this!

lencodes (changes requested):
Looks good! Couple of feedback points.

PR #196, opened 2022-03-04
Formio modal improvement

bbrands02:


pwa/src/components/objectEntities/ObjectEntitiesTable.tsx
```diff
@@ -161,7 +161,16 @@ const ObjectEntitiesTable: React.FC<ObjectEntitiesTableProps> = ({ entityId }) =
             <Modal
               title={`Create a new ${entity?.name} object`}
               id="objectModal"
-              body={() => <>{FormIO && FormIO}</>}
+              body={() => (
+                <>
+                  { FormIO && formIOSchema.components && FormIO }
```
lencodes, line 166: I think we could make sure (or maybe already are making sure..) that `formIOSchema.components` is available, so we don't have to do this check in the render function.

1. Can `FormIO` exist without it?
2. We're checking on line `42` for `formIOSchema` before setting `FormIO`, this check might already be enough, if not, we might want to change this check to `formIOSchema.components` and
3. The check could then be `{ FormIO && FormIO }` or maybe just `{ FormIO }` without a check.

pwa/src/components/objectEntities/ObjectEntitiesTable.tsx
```diff
@@ -161,7 +161,16 @@ const ObjectEntitiesTable: React.FC<ObjectEntitiesTableProps> = ({ entityId }) =
...
+                  {
+                    entity?.handlers?.length < 1 && (
```
lencodes, line 168: This is basically a [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy?retiredLocale=nl) check. Right now, you're checking on `< 0`, in reality it would never be `-1`, it's minimum would be `0`.

Therefore, the check would more logically be `entity?.handlers?.length === 0 &&`. But because `0` is falsy, we can treat it as such (something like how we would treat a boolean), like so:

```
!entity?.handlers?.length && <>...</>
```

lencodes (changes requested):

PR #200, opened 2022-03-07
Selectboxen selecteeren bug

hghanbari:


pwa/src/components/endpoints/endpointForm.tsx
```diff
@@ -191,7 +191,7 @@ export const EndpointForm: React.FC<EndpointFormProps> = ({ endpointId }) => {
                         {applications ? (
                           <SelectInputComponent
                             options={applications}
-                            data={endpoint?.application?.name}
+                            data={endpoint?.applications?.id}
```
lencodes: This doesn't work, this has to be `.name`.

I think we made this change during our call and tested it too hastily.

Works fine as `endpoint?.applications?.name`

pwa/src/components/entities/entityForm.tsx
```diff
@@ -236,7 +236,7 @@ export const EntityForm: React.FC<EntityFormProps> = ({ entityId }) => {
                             {entity !== null && entity.gateway !== undefined && entity.gateway !== null ? (
                               <SelectInputComponent
                                 options={sources}
-                                data={entity.gateway.name}
+                                data={entity.gateway.id}
```
lencodes: This doesn't work. The old implementation didn't work either. 

The reason behind this is that the `SelectInputComponent` is a bit complex to grasp, but the fix is as followed:

1. The `options` prop we're sending needs a valid `name` and `value` for the component to work and then
2. We need to set the current value in the `value` prop (which has to be the same as `option.value` here to be selected by default..

So here's how we do that:

1. We need to `.map()` the sources that we `GET` from the backend, like so:

```
const _sources = res.data.map((source) => ({
  name: source.name,
  value: `/admin/attributes/${source.id}`,
  }));
setSources(_sources);

//NOTE: this was just setSources(res.data); but now we're mapping the data to the structure we need in the selectelement!
```

2. We need to update the `SelectInputComponent` implementation, like so:

```
<SelectInputComponent
                            options={sources}
                            data={`/admin/attributes/${entity?.gateway?.id}`}
                            name={"gateway"}
                            id={"gatewayInput"}
                            nameOverride={"Source"}
                          />
```

pwa/src/components/objectEntities/ObjectEntityForm.tsx
```diff
@@ -237,7 +237,7 @@ export const ObjectEntityForm: React.FC<ObjectEntityFormProps> = ({ objectId, en
                               objectEntity.application !== null ? (
                                 <SelectInputComponent
                                   options={applications}
-                                  data={objectEntity.application.name}
+                                  data={objectEntity.applications.id}
```
lencodes: This entire file is never included anywhere, so I feel as if this change did nothing (and was not tested, I assume).

lencodes (changes requested):
I left some comments, I feel as if this could've been tested better before submitting for review.

PR #202, opened 2022-03-09
session table

smisidjan:
added a table for sessionValues.

 If it's important to show all data in session values, a table might not be the best solution. I think the codeBlock was better to show all the data.

### DIFF
--- a/pwa/src/components/logs/logModal/LogModal.tsx
+++ b/pwa/src/components/logs/logModal/LogModal.tsx
@@ -17,7 +17,7 @@ const LogModal: React.FC<LogModalProps> = ({ log }) => {
         key={log.id}
         title={"Call log"}
         id={`logs${log.id}`}
-        body={function () {
+        body={function() {
           const statusClass = log.responseStatusCode ? log.responseStatusCode > 199 && log.responseStatusCode < 300 ? "success" : "danger" : "danger";
           return (
             <>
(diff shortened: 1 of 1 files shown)


bbrands02 (changes requested):
I think its better to loop through sessionValues with a Object().entries function and create a table row each iteration because whats in sessionValues is not the same for each Log

pwa/src/components/logs/logModal/LogModal.tsx
```diff
@@ -100,19 +100,35 @@ const LogModal: React.FC<LogModalProps> = ({ log }) => {
...
+                                  </tr>
+                                  <tr>
+                                    <th>Organizations</th>
+                                    <td>{log.sessionValues.organizations[0]}</td>
+                                  </tr>
+                                  <tr>
+                                    <th>Application</th>
+                                    <td>{log.sessionValues.application.name}</td>
+                                  </tr>
+                                </table>
```
lencodes, line 124: (I think the following feedback will also resolve your comment about having to show all content in the description of this pull request)

Right now, you're hardcoding the rows (_Active organisation, Parten organisation, Organisations_ and _Application_). This works fine for these rows, however, when a session has more values it will never show them (e.g.  `callId`, `handler`, `endpoint`, ...).

A better solution might be (the same as we implemented in the _Headers_ accordions:

```
const logs = [];
for (const [key, value] of Object.entries(log.requestHeaders)) {
  logs.push({ ...{ key, value } });
}

// at this point we have a mappable array (logs) with simpele {log} entries
// so now we're able to map over the logs and render them, like so:

{logs.map((log, idx) => {
  return (
    <tr key={log.key + idx}>
    <th>{log.key}</th>
    <td>{log.value}</td>
    </tr>
   );
})}
```

Note: some of these `{ log.value }` entries are quite big (they are code objects themselves). So it might be best to show them as `CodeBlocks`. Maybe you can verify this with @rubenvdlinde. I think just rendering these big objects or arrays like plain text wouldn't be a good idea.

lencodes (changes requested):

PR #203, opened 2022-03-09
Resource Delete Button

remko48:


pwa/src/components/deleteModal/DeleteModal.tsx
```diff
@@ -0,0 +1,49 @@
+import * as React from "react";
+import "./deletemodal.css";
+import { Modal } from "@conductionnl/nl-design-system";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faTrash } from "@fortawesome/free-solid-svg-icons";
+
+interface DeleteModalProps {
+  handleDelete: any;
```
lencodes: This `prop` shouldn't be `:any`, I think it's always an function (and probably always a `void` function). So that would become something like:

`handleDelte: () => void`

pwa/src/components/deleteModal/DeleteModal.tsx
```diff
@@ -0,0 +1,49 @@
...
+  handleId: any;
```
lencodes: I would recommend renaming this.

1. I'm unsure if you mean a _handler_ here (as in the resource coupled to endpoints), if you do, this would be way too specific because this delete modal has nothing to do with the handlers or
2. If you meant it as something to _handle_ (like _handleDelete_ handles the deletion of something): this is not actually handling anything (it's not a function).

Therefore, I would rename it to something like `resourceId`.

lencodes: Edit: and it should't be `:any`.

pwa/src/components/deleteModal/DeleteModal.tsx
```diff
@@ -0,0 +1,49 @@
...
+  optionalLength?: any;
```
lencodes: Very nice catch, that we (might or probably should) want to give the user feedback if he's deleting more than just the resource that he's clicking on.

However, I think we could probably do this in a more generic way (see other comments regarding the `optionalLength`).

Note: this shouldn't be `:any`.

Note 2: you're actually checking if it's `=== 1`, this would break on a string.

pwa/src/components/deleteModal/DeleteModal.tsx
```diff
@@ -0,0 +1,49 @@
...
+          return (
+            <>
+              {optionalLength === 1 ? (
+                <p>
+                  Are you sure you want to delete {handleId}? <br></br>
+                  <b>If you delete this translation you wil also delete the Translation Table!</b>
+                </p>
+              ) : (
+                <p>Are you sure you want to delete {handleId}?</p>
+              )}
```
lencodes: We're now duplicating the `<p>Are you sure you want to delete {handleId}?</p>` part. An easier (and better readable) way of doing this would be:

```
<p>
  Are you sure you want to delete {handleId}? <br></br>
  { optionalLength === 1 && <b>If you delete this translation you wil also delete the Translation Table!</b> }
</p>
```

This halves the amount of code we have to write, and removes a hard to read (and hard to maintain) ternary (`? :`) statement.

pwa/src/components/deleteModal/DeleteModal.tsx
```diff
@@ -0,0 +1,49 @@
...
+    <div className="DeleteModal">
+      <Modal
+        title="Warning!"
+        id={`deleteModal${handleId}`}
+        body={function () {
+          return (
+            <>
+              {optionalLength === 1 ? (
+                <p>
+                  Are you sure you want to delete {handleId}? <br></br>
```
lencodes: You're using `<br/>` tags here to add a form of styling. I would recommend adding this styling to the `deleteModal.css`.

pwa/src/components/deleteModal/DeleteModal.tsx
```diff
@@ -0,0 +1,49 @@
...
+                  <b>If you delete this translation you wil also delete the Translation Table!</b>
```
lencodes: This is an important one to understand (if something's unclear about this comment, call me, we can go over it 👍🏼 )

Right now, you're adding logic within the `DeleteModal` based on where it is implemented (`TranslationTable`). So, you're writing logic for outside context. A couple of issues arise here:

1. The `DeleteModal` gets more complexity (especially for maintenance, you understand why you did this, other people might not);
2. When we have another edge case, we have to add more of this complex logic to the `DeleteModal`.

What we want is for any component (in this case the `DeleteModal`) so be completely unaware of its context (so the `DeleteModal` shouldn't care that it's being implemented inside a `TranslationTable`).

lencodes: I think a possible solution in this situation would be (if you have another idea, feel free):

1. Change the `optionalLength` prop to something like `additionalMessage?: string` and
2. Now you can just print this additional message if it's set.

This way, the additional message would be set **within** the `TranslationTable` and therefore the `DeleteModal` would not need any changes if different implementations require different messages.

pwa/src/components/deleteModal/DeleteModal.tsx
```diff
@@ -0,0 +1,49 @@
...
+              <button
+                className="utrecht-button btn-sm btn-danger mr-2"
+                onClick={() => handleDelete(handleId)}
+                data-bs-dismiss="modal"
+              >
+                <FontAwesomeIcon icon={faTrash} /> Delete
+              </button>
+              <button className="utrecht-button cancel-btn-right btn-sm mr-2" data-bs-dismiss="modal">
+                Cancel
+              </button>
```
lencodes: For me, it felt unnatural to have the delete button on the left, I would swap them.

pwa/src/components/deleteModal/deletemodal.css
```diff
@@ -0,0 +1,3 @@
+.cancel-btn-right {
+  float: right;
+}
```
lencodes: Two things here: 

1. I would recommend working with `flex`, if you're unsure how this works [check out this guide](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) and
2. Your `DeleteModal` is wrapped in a `.DeleteModal` className, we should use this to style everything within this component, like so:

```
.DeleteModal .DeleteModal-button--cancel {
  ... 
}
```

This way, the `css` logic written in this file, can never unintentionally target elements outside of this component.

lencodes (changes requested):
Functionally it looks great 👍🏼 

I did leave some comments on how we can make it a bit better maintainable.

pwa/src/components/applications/applicationsTable.tsx
```diff
@@ -59,17 +62,19 @@ export default function ApplicationsTable() {
...
-        })
-        .catch((err) => {
-          setAlert({ title: "Oops something went wrong", message: err, type: "danger" });
-          throw new Error("DELETE application error: " + err);
-        });
-    }
+    setLoadingOverlay(true);
+    API.Application.delete(id)
+      .then(() => {
+        setAlert({ message: `Deleted application`, type: "success" });
```
lencodes: Whenever you're just passing a `string` as an argument (in the case of `message` here). You do not have to use backticks (`) but you can (and should) just use quotes (").

pwa/src/components/applications/applicationsTable.tsx
```diff
@@ -59,17 +62,19 @@ export default function ApplicationsTable() {
...
+        handleSetApplications();
+      })
+      .catch((err) => {
+        setAlert({ message: err, type: "danger" });
```
lencodes: You are (unintentionally I assume) overwriting some changes made to this `setAlert()`. The previous one had a title and a message instead of only the message.

lencodes: You are (unintentionally I assume) overwriting some changes made to this `setAlert()`. The previous one had a title and a message instead of only the message.

pwa/src/components/applications/applicationsTable.tsx
lencodes: Check this throughout your updated files please.

lencodes: Check this throughout your updated files please.

pwa/src/components/applications/applicationsTable.tsx
lencodes: Check this throughout your updated files please.

lencodes: Check this throughout your updated files please.

pwa/src/components/deleteModal/DeleteModal.tsx
```diff
@@ -0,0 +1,46 @@
...
+  optionalMessage?: JSX.Element;
+}
+
+const DeleteModal: React.FC<DeleteModalProps> = ({ resourceDelete, resourceId, optionalMessage }) => {
+  return (
+    <div className="DeleteModal">
+      <Modal
+        title="Warning!"
+        id={`deleteModal${resourceId}`}
+        body={function () {
```
lencodes: You can shorthand this notation making it easier to read and maintain. 

Just like on line `8` where we're declaring `resourceDelete: () => void` which basically means a function that returns a void. We can do the following here:

```
() => <> component here </>
```

This works a follwed:

```
() => {
  <>component</>
  // this is not instantly returned, because it's within brackets ({})
}

() => <>component</> // this IS instantly returned, because it is NOT in brackets

() => (
  <>component</> // this IS ALSO instantly returned, because it is NOT in brackets
)
```

lencodes: You can shorthand this notation making it easier to read and maintain. 

Just like on line `8` where we're declaring `resourceDelete: () => void` which basically means a function that returns a void. We can do the following here:

```
() => <> component here </>
```

This works a follwed:

```
() => {
  <>component</>
  // this is not instantly returned, because it's within brackets ({})
}

() => <>component</> // this IS instantly returned, because it is NOT in brackets

() => (
  <>component</> // this IS ALSO instantly returned, because it is NOT in brackets
)
```

lencodes (changes requested):
Looks good 👍🏼 

Few really small changes to make, you can merge it afterwards. If you have any questions, you can call me.

PR #208, opened 2022-03-09
added extra if statement to prevent infinite loop

remko48:


lencodes (approved):
Looks good 👍🏼 

(I do think we should think about a better solution for these things, it's becoming quite hard to manage with all those if statements)

(Maybe we should create some gateway endpoints that return everything the frontend needs, instead of having to do multiple calls).

PR #209, opened 2022-03-09
GW-510 and GW-511

smisidjan:
* added apiCalls with filter on type
* filtert dashboard logTable on incoming calls
* added outgoing tab
* added table with outgoing calls from callId


smisidjan:
I added a check for showing the modal. 
If we want to open an other modal (outgoing log) on top of the incoming log modal we need to tink how we want to do that. I make an issue for that so this issue can be closed

bbrands02 (approved):

PR #215, opened 2022-03-10
Feature/gw 517/logs table refresh

lencodes:
This pull request also includes the "Help" button from GW-516.


bbrands02 (approved):
Getest en approved

PR #216, opened 2022-03-10
Feature/gw 20/expired jwt redirect

lencodes:
This pull request includes:

1. Addition of `JWT` expiration validation to `gatsby-router` (which checks every time a route changes, so on every page change);
2. Addition of a generic `Send()` method within the `APIService` which is now implemented in every `resource`;
3. Addition of `JWT` expiration validation in the `Send()` method.

This results in a redirect to `/login` whenever the `JWT` expires.


bbrands02 (approved):
Tested by removing jwt from session storage in browser, not tested by waiting for jwt token to expire, but we will probably experience that by just working on the admin ui

PR #217, opened 2022-03-10
ObjectEntity config accordion and create object page

bbrands02:


pwa/src/components/objectEntities/objectEntityFormNew/ObjectEntityFormNew.css
```diff
@@ -0,0 +1,17 @@
...
+    padding: 7px 5px;
+}
+
+.configurationTable-table {
+    width: 100%;
+}
+
+.configurationTable-table tr th {
+    width: 30%;
+}
```
lencodes: This should, in this case, all be prefixed with `.ObjectEntityFormNew-`. This way, these css rules never unintentionally target anything outside of this component.

pwa/src/components/objectEntities/objectEntityFormNew/ObjectEntityFormNew.tsx
```diff
@@ -0,0 +1,262 @@
+import * as React from "react";
```
lencodes: This should not be named `..New`. Can we either remove the old implementation and rename this _or_ give this another name?

pwa/src/components/objectEntities/objectEntityFormNew/ObjectEntityFormNew.tsx
```diff
@@ -0,0 +1,262 @@
...
+                                },
+                              },
+                              {
+                                title: "External Result",
+                                id: "externalResultAccordion",
+                                render: function () {
+                                  return (<h6>IN PROGRESS: needs refinement, no info on how to show data</h6>);
+                                },
+                              },
+                            ]}
```
lencodes: All of this is dummy data (see top-level comment of this review).

pwa/src/components/objectEntities/objectEntityFormNew/ObjectEntityFormNew.tsx
```diff
@@ -0,0 +1,262 @@
...
+                                id: "errorsAccordion",
+                                render: function () {
+                                  return (<h6>IN PROGRESS: needs refinement, no info on how to show data</h6>);
+                                },
+                              },
+                              {
+                                title: "Promises",
+                                id: "promisesAccordion",
+                                render: function () {
+                                  return (<h6>IN PROGRESS: needs refinement, no info on how to show data</h6>);
```
lencodes: I don't think this should be a `<h>` tag. It's usually used to bring structure (and hierarchy) to e.g. articles. And when they're used, they should have a clear hierarchy (so top-level `1`, second-level `2`, and so on).

However, because this is not any kind of article, it's probably best to just keep this as a `<span>` or something along those lines. 

Also, regarding the hierarchy: this `6` is now rendered directly below a `2` (the title of the accordion).

lencodes (changes requested):
I would like to re-review this when it's further along.

I think we should try to fix the following things (besides the inline comments I left) before merging:

1. Make this actually functional (so no dummy data), right now we cannot actually configure anything within the _advanced configuration_ (also, we do not have the accordion while creating a new object) and
2. The _advanced configuration_ should be **inside** the form, not below it.

Let me know if this feedback is realistic.

Note: maybe it would've been better to start with the backend changes instead of with the frontend ones?

bbrands02:
Testable by pulling the new gateway image and viewing/creating objects from the object type 'weer'

PR #218, opened 2022-03-11
LogModal show source if type out

bbrands02:


lencodes (changes requested):
This is based on a `development` branch that had an infinite loop (added on Wednesday). This loop is created in combination of the `LogModal` and `CodeBlock`.

Can you please merge development into this branch (which should fix the loop) and then re-request a review?

PR #219, opened 2022-03-11
fix source form

smisidjan:
* fix error bij type soap
* small refactor on the inputs of the form


lencodes (approved):
Seems to work again.

Nice refactor of the rest of the file, much better this way 👍🏼

PR #220, opened 2022-03-11
Bugfix/gw 124/errors and warnings

lencodes:
See commit messages.


bbrands02 (changes requested):
Nice, no console errors except one but that won't impediment the approval. 
image

There is a console.log in LogModal.tsx:51 that should be removed

PR #222, opened 2022-03-14
GW-494 Checkboxes subscriber form

smisidjan:
subscriber form modified using the attribute form


pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -57,57 +57,52 @@ export const SubscriberForm: React.FC<SubscriberFormProps> = ({ subscriberId, en
...
+  React.useEffect(() => {
+    if (subscriberId) {
+
+      if (( !subscriber || !sources || !endpoints || !tableNames) && !showSpinner) setShowSpinner(true);
+      if (( subscriber && sources && endpoints && tableNames) && showSpinner) setShowSpinner(false);
+    } else {
+
+      if (( !sources || !endpoints || !tableNames) && !showSpinner) setShowSpinner(true);
+      if (( sources && endpoints && tableNames) && showSpinner) setShowSpinner(false);
+    }
```
lencodes: This is becoming a very complex check (for me it was quite hard to understand what exactly is going on).

I think think we're doing two things within this check that could be done easier:

1. We're completely duplicating the check based on `subscriberId` (and when it's available we have to add the `subscriber` check) and
2. We're completely duplicating the inverted check (so setting to `true` or `false`).

So updating this check in order of above points:

1. Would result in something like:

```
if (!sources || !endpoints || !tableNames || (subscriberId && !subscriber)) {
  setShowSpinner(true);
  return;
}

setShowSpinner(false);
```

2. Also adding this point would end up something like:

`setShowSpinner(!sources || !endpoints || !tableNames || (subscriberId && !subscriber));` <= I think only this line would do the trick

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -57,57 +57,52 @@ export const SubscriberForm: React.FC<SubscriberFormProps> = ({ subscriberId, en
...
+  }, [subscriber, sources, endpoints, tableNames]);
```
lencodes: This dependency array should also contain `subscriberId` (because when it changes, the logic within this effect changes).

lencodes (changes requested):
Nice refactor 👍🏼 

Left some small suggestions.

PR #225, opened 2022-03-14
added spacing 

smisidjan:


lencodes (changes requested):
Even though `<br>` elements create spacing, they are meant only to break a line in a block of text. And they are therefore **not meant to space layouts**.

This spacing should be created with `css` rules.

PR #227, opened 2022-03-14
GW-539 Refactor alert title

smisidjan:
See commit comments for more context


pwa/src/components/alert/alert.tsx
```diff
@@ -15,8 +15,8 @@ const AlertComponent = () => {
     return (
       <div className={`conduction-alert conduction-alert-${alert.type}`}>
         {
-          alert.title && (
-            <h5>{alert.title}</h5>
+          alert.type === 'danger' && (
+            <h5><span>Oops something went wrong</span></h5>
```
lencodes: Because this is not a `sub-title` of a (e.g.) article or some other hierarchal form of content, it doesn't have to be a `h5` tag (or any other `h` tag). These are reserved to bring hierarchy into content and this pop-up is not part of that (or any) hierarchy.

Resulting in that this can just be a `<span>` without anything else.

Note: this could mean that we have to write some `css` to style it.

lencodes: Also, right now we're writing this syntax pretty complex, which I think is unnecessary. This would do the same (and would comply with out prettier):

```
{alert.type === "danger" && <span>Oops something went wrong</span>}
```

lencodes: Also, right now we're writing this syntax pretty complex, which I think is unnecessary. This would do the same (and would comply with out prettier):

```
{alert.type === "danger" && <span>Oops something went wrong</span>}
```

pwa/src/components/alert/alert.tsx
```diff
@@ -15,8 +15,8 @@ const AlertComponent = () => {
...
           )
         }
         <p>{alert.message}</p>
```
lencodes: Here goes the same feedback about hierarchy. This is not a paragraph and should therefore not be wrapped in a `p` tag. Here we can also just use a `span` tag.

lencodes: Here goes the same feedback about hierarchy. This is not a paragraph and should therefore not be wrapped in a `p` tag. Here we can also just use a `span` tag.

lencodes (changes requested):
Nice 👍🏼 

See some inline comments for a few small feedback notes.

One more thing: I don't think your `.prettierrc` is implemented correctly. On line `18` you're using single quotes and you're missing a semicolon on line `20` both of which should never happen:


```
"parser": "typescript",
                "printWidth": 120,
                "trailingComma": "all", <====
                "tabWidth": 2,
                "singleQuote": false <=====
```

PR #229, opened 2022-03-15
Feature/gw 520/design decisions

lencodes:


MWest2020 (changes requested):
All forms are built upon [React Hook Form](https://react-hook-form.com/) because of its small bundle size, performance, and, most importantly, its automatic validation of form-data. Combined, React Hook Form, removes complexity from our codebase and creates a more robust, secure, and efficient solution.

-   All asynchronous calls are realized upon [React Query](https://react-query.tanstack.com/) because of its automatic request caching, automatic re-trying mechanisms, automatic request validations, and the ability to invalidate requests. Combined, React Query creates a more robust and efficient solution _and_ removes the need for a "global state" manager.


Spell /Grammar check

apiService.md
```diff
@@ -103,3 +103,7 @@ API.Attribute.getAll()
 ```
 
 > This is an example of everything implemented
+
+## Automatic session validation
+
```
MWest2020, line 108: ## Automatic session validation

With the `APIService` generic `Send()` method, we automatically validate the user's session **before** fulfilling a send request. This way, invalid requests are never sent. In the case of an invalid request, the user logs out automatically.

MWest2020: "its" removed (meant to be " 's"?) and altered last sentence (was in passive voice)

MWest2020 (changes requested):
Overal nice. Great to see more additions to the documentation

PR #231, opened 2022-03-15
LogTable column refactor

hghanbari:


lencodes (changes requested):
Ticket GW-553 requires three changes, in this pull request I only see one.

lencodes (changes requested):
Hey @hghanbari, this pull request includes a lot of unintentional changes. So this pull request needs to be updated. 

Also, you're making some changes in forms that have been heavily refactored in the previous days, this is going to create a lot of merge conflicts. I recommend **only making changes in the required files** instead of these small changes in forms as well.

If you have any questions, you can call me.

PR #232, opened 2022-03-15
MultiSelect aanpasbare title

hghanbari:


pwa/src/components/common/elementCreationNew.tsx
```diff
@@ -75,7 +75,7 @@ const ElementCreationNew: React.FC<ElementCreationNewProps> = ({ id, label, data
       )}
 
       <div className="row">
-        <h5>Create {label}</h5>
+        <span className="font-weight-bold">Create {label}</span>
```
lencodes, line 78: The input placeholder underneath says _Add_ instead of _Create_, making it look a bit odd. Maybe change the placeholder to _Create_ as well?

image

lencodes (changes requested):
Looks good to me, left one small comment.

PR #233, opened 2022-03-15
refactored TextereaGroup label is now required and can be a JSX.Element and added link to json logic

remko48:


lencodes (approved):
Looks good 👍🏼

PR #236, opened 2022-03-16
bug voor URL input

hghanbari:


pwa/src/components/sources/sourceForm.tsx
```diff
@@ -22,6 +22,7 @@ import LoadingOverlay from "../loadingOverlay/loadingOverlay";
 import { AlertContext } from "../../context/alertContext";
 import { HeaderContext } from "../../context/headerContext";
 import MultiDimensionalArrayInput from "../common/multiDimensionalArrayInput";
+import { validateJSON } from "../../services/validateJSON";
```
lencodes, line 25: This is unused.

pwa/src/components/sources/sourceForm.tsx
```diff
@@ -318,7 +319,7 @@ export const SourceForm: React.FC<SourceFormProps> = ({ sourceId }) => {
                       </div>
                       <div className="col-6">
                         <GenericInputComponent
-                          type={"text"}
+                          type={"password"}
```
lencodes, line 322: When you change an `input` to `type=password` you cannot see what you're typing anymore. Since this is not described in the ticket, we will have to verify this.

To do: check with @rubenvdlinde if changing this input to `type=password` is wanted.

lencodes (changes requested):
Good solution, couple of small comments.

PR #237, opened 2022-03-16
Maxlengte limit voor Route

hghanbari:


lencodes (changes requested):
Couple of things to check and fix with this pull request **before** merging it.

1. Did you test this? I think this might still not work.

2. Currently, I'm getting the following error when trying to send the form (therefore, I cannot test the form):

image

3. You're setting the `maxLength` to `1406`. I'm assuming you've based this on the `1406` error message in the ticket. However, this `1406` is not the expected length of the field. `1406` [is an error number from MySQL](https://tableplus.com/blog/2019/09/data-too-long-for-column.html#:~:text=That%20error%20message,the%20column%20size.).

I think you tried to fix this too hastily.

lencodes (changes requested):
I don't think you understood my feedback, let's call about this ticket.

hghanbari:
> I don't think you understood my feedback, let's call about this ticket.

PR #241, opened 2022-03-18
Feature/gw 522/endpoints react query

lencodes:
This pull request includes:

- Addition of `react-query`
- Implementation of `react-query` in `Endpoints` (and all forms that require `endpoints`)

It also includes the following details:

- `staleTime` is set to 5 minutes (meaning calls are valid for five minutes, except for when we invalidate them ourselves)
- Optimistic update, create and delete is set on resource details
- Optimistic GET is set on resource details (they are grabbed from the list query)

Note 1: To get optimistic deletion to work, I had to update the `DeleteModal` and its implementations and its linked `APIService` methods.

Note 2: The collections form seems to be broken (I think it already was broken before this update).

### DIFF
--- a/pwa/src/components/endpoints/endpointsTable.tsx
+++ b/pwa/src/components/endpoints/endpointsTable.tsx
@@ -17,25 +17,32 @@ export default function EndpointsTable() {
   const [__, setHeader] = React.useContext(HeaderContext);
 
   const queryClient = useQueryClient();
-  const getEndpointsQuery = useQuery<any[], Error>("endpoints", API.Endpoint.getAll);
-  const deleteEndpointMutation = useMutation<any, Error, any>(API.Endpoint.delete);
+
+  const getEndpointsQuery = useQuery<any[], Error>("endpoints", API.Endpoint.getAll, {
+    onError: (error) => {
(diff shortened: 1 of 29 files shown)


pwa/src/components/endpoints/endpointsTable.tsx
```diff
@@ -8,43 +8,41 @@ import { HeaderContext } from "../../context/headerContext";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faTrash, faEdit } from "@fortawesome/free-solid-svg-icons";
 import DeleteModal from "../deleteModal/DeleteModal";
-import LoadingOverlay from "../loadingOverlay/loadingOverlay";
```
remko48, line 11: You have deleted the loadingOverlay wich means that when you delete an item you will be able to edit that item you are about to delete.

remko48 (changes requested):

pwa/src/components/endpoints/endpointsTable.tsx
lencodes: Nice catch. Fixed.

PR #243, opened 2022-03-18
form verzenden refactor

hghanbari:


lencodes (changes requested):
The way you've changed it in `sourceForm.tsx` is correct. However, you have now done it **only** for the `sources` resource. With _resources_ we meant all available resources, so: `sources`, `entities`, `object_entities`, `handlers`, ...

When you've updated everything, you should also re-update the `design.md` to reflect the changes 👍🏼

pwa/design.md
```diff
@@ -1,5 +1,7 @@
 # Design Decisions
 
+After user submits a new "source" form, they are redirected to the edit form for the created source entry.
+
 ## Introduction:
 
 The following document contains all design decisions. Any and all included packages **must be** _highly adopted_ and _well documented_.
```
lencodes: Right now, you're adding this decision at the top of the document. However, there's a section called _Decision log_, in there you can make another entry with this decision.

Also, this is now only describing the `source`, even though these redirects are way more generic.

pwa/src/components/applications/applicationForm.tsx
```diff
@@ -135,9 +135,13 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({ id }) => {
     }
 
     API.Application.createOrUpdate(body, id)
-      .then(() => {
+      .then((response) => {
```
lencodes: Goes for all changed files:

Throughout the application we've been using `res` instead of `response`. I'd recommend updating this.

pwa/src/components/applications/applicationForm.tsx
```diff
@@ -135,9 +135,13 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({ id }) => {
...
         setAlert({ message: `${id ? "Updated" : "Created"} application`, type: "success" });
-        navigate("/applications");
+        if (id) {
+          navigate(`/applications`);
+        } else {
+          navigate(`/applications/${response.data.id}`);
+        }
```
lencodes: Goes for all changed files:

I'm thinking that the following syntax might be easier to read:

```
id && navigate("/applications");
!id && navigate(`/applications/${res.data.id}`);
```

Also note: you have currently used backticks in the first `navigate`, since this does not contain any javascript you can just use regular quotes (") (like in my example).

lencodes (changes requested):

pwa/design.md
```diff
@@ -1,5 +1,7 @@
 # Design Decisions
 
+After user submits a new form, they are redirected to the edit form for the created source entry.
```
lencodes, line 3: Better. However, I do have some feedback.

There is **already** a section called _Decision Log_, this is where this should be placed. So underneath the `React Hook Form` and `React Query` decisions, you can add the following decision:

- Form redirects throughout the application are based on the action a user takes. On a _create_, the user is always redirected to the detail page of the created resource, on an _update_ the user is always redirected back to the collection page.

pwa/src/components/handlers/handlerForm.tsx
```diff
@@ -142,9 +142,13 @@ export const HandlerForm: React.FC<HandlerFormProps> = ({ handlerId, endpointId
     }
 
     API.Handler.createOrUpdate(body, handlerId)
-      .then(() => {
+      .then((response) => {
```
lencodes, line 145: This still says `response`. Same in many other files.

lencodes (changes requested):
Better, however I think you did not fully understand my feedback. Let's call about this ticket.

PR #248, opened 2022-03-24
Feature/gw 612/select element

lencodes:
This pull request includes:

- Addition of a new `SelectSingle` component;
- Some small changes to the retrieval of data to correctly map the `applications` for the (multi-)select elements


remko48 (approved):
Looks good.
Only small comment, when you have saved an application to an endpoint and the name of the application changes it wil not change the name of the already selected application. After saving it again it will change.

lencodes:
> Looks good.
> Only small comment, when you have saved an application to an endpoint and the name of the application changes it wil not change the name of the already selected application. After saving it again it will change.

Nice catch. I think this will resolve itself once we've implemented `react-query` for applications as well. 👍🏼

PR #254, opened 2022-03-25
Feature/gw 606/react hook form source

lencodes:
This pull request includes:

- Integration of `react-hook-form` into `SourceForm`;
- Implementation of `CreateArray` and `CreateKeyValue` components into `SourceForm`;
- Addition of `InputUrl` component for `type="url"` input components.

### DIFF
--- a/pwa/src/components/sources/sourceForm.tsx
+++ b/pwa/src/components/sources/sourceForm.tsx
@@ -59,7 +59,6 @@ export const SourceForm: React.FC<SourceFormProps> = ({ sourceId }) => {
     data.type = data.type && data.type.value;
     data.auth = data.auth && data.auth.value;
 
-    console.log(data);
     API.Source.createOrUpdate(data, sourceId)
       .then(() => {
         setAlert({ type: "success", message: `${sourceId ? "Updated" : "Created"} source` });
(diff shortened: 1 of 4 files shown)


pwa/src/components/sources/sourceForm.tsx
```diff
@@ -37,25 +25,98 @@ export const SourceForm: React.FC<SourceFormProps> = ({ sourceId }) => {
...
+      .then(() => {
+        setAlert({ type: "success", message: `${sourceId ? "Updated" : "Created"} source` });
+        navigate("/sources");
+      })
+      .catch((err) => {
+        setAlert({ type: "danger", message: err.message });
+        throw new Error("Create or update source error: " + err);
+      })
+      .finally(() => {
+        setLoadingOverlay(false);
```
remko48, line 74: You are setting the LoadingOverlay to false but you forgot the line of `setLoadingOverlay(true);` on line 59 so the spinner is never shown.

remko48 (changes requested):
Looks good only one small comment

pwa/src/components/sources/sourceForm.tsx
lencodes: Thanks!

PR #257, opened 2022-03-25
REFACTOR collection form to react-hook-form

lencodes:
**Note:** I'm aware that the `GitHub URL` field is not saving correctly. I don't think it did _before_ this refactor and I don't think I can fix this without backend changes. Therefore, I've made [this ticket](https://conduction.atlassian.net/browse/GW-617).


bbrands02:
image
 Seems like the endpoints select is showing some weird behaviour

bbrands02 (changes requested):
^

PR #259, opened 2022-03-28
EYF-611, 610. 583, 582

smisidjan:


pwa/src/apiService/apiService.ts
```diff
@@ -16,6 +16,7 @@ import ApiCalls from "./resources/apiCalls";
 import Subscriber from "./resources/subscriber";
 import Collection from "./resources/collection";
 import { isLoggedIn, logout, validateSession } from "../services/auth";
```
lencodes: `isLoggedIn` is no longer used and can therefore be removed as a dependency.

pwa/src/apiService/apiService.ts
```diff
@@ -16,6 +16,7 @@ import ApiCalls from "./resources/apiCalls";
...
+import Repository from "./resources/repository";
```
lencodes, line 18: All resources are currently imported above the services, might be nice to move this import one line up, above the services.

pwa/src/templates/dashboard/Dashboard.tsx
```diff
@@ -100,6 +102,15 @@ const Dashboard: React.FC = () => {
...
           />
+
+          <DashboardCard
+            amount={getRepositoriesQuery.isSuccess ? getRepositoriesQuery.data.length : 0}
+            title="Github Shop"
+            iconBackgroundColor="5bc0de"
+            icon={<i className="fas fa-shopping-cart fa-lg"/>}
+            subtitle="View all github repositories"
+            linkTo="shop"
+          />
```
lencodes: I'm getting the following error on first load:

image

lencodes: Your `Repository.getAll()` is returning `undefined`. Something in the call itself is failing:

image

lencodes (changes requested):
The code is breaking, I've left a comment on where it is breaking. Can you please fix this and then request a re-review?

PR #260, opened 2022-03-29
Feature/gw 619/generic tooltip

lencodes:
Read commit messages for made changes.

How to implement:

In any input component (e.g. `<InputText />`) you can now pass the `tooltipContent` prop to automatically create a tooltip next to the label, like so:

```
<InputText
  label="Name"
  name="name"
  tooltipContent="Content goes here" <== string implementation
  {...{ register, errors }}
  validation={{ required: true }}
/>
```

```
<InputText
  label="Name"
  name="name"
  tooltipContent={<>Content goes here</>} <== JSX.Element implementation
  {...{ register, errors }}
  validation={{ required: true }}
/>
```

### DIFF
--- a/pwa/src/components/tooltip/tooltip.css
+++ b/pwa/src/components/tooltip/tooltip.css
@@ -1,0 +1,13 @@
+.Tooltip > .Tooltip-button {
+  cursor: pointer;
+  color: var(--utrecht-topnav-list-background-color);
+}
+
+.Tooltip > .Tooltip-tooltip {
+  pointer-events: auto !important;
+}
+
+.Tooltip > .Tooltip-tooltip:hover {
+  visibility: visible !important;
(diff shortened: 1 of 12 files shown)


pwa/src/components/tooltip/tooltip.css
```diff
@@ -0,0 +1,13 @@
...
+}
+
+.Tooltip > .Tooltip-tooltip {
+  pointer-events: auto !important;
+}
+
+.Tooltip > .Tooltip-tooltip:hover {
+  visibility: visible !important;
+  opacity: 1 !important;
+}
```
lencodes, line 13: The `!important` keywords are suggested by `react-tooltip` themselves.

bbrands02 (approved):

PR #262, opened 2022-03-30
Feature/gw 607/react form subsciber

bbrands02:


pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -104,69 +120,75 @@ export const SubscriberForm: React.FC<SubscriberFormProps> = ({ subscriberId, en
...
-        setAlert({ message: `${subscriberId ? "Updated" : "Created"} subscriber`, type: "success" });
+        setAlert({ type: "success", message: `${subscriberId ? "Updated" : "Created"} subscriber` });
         navigate(`/entities/${entityId}`, {
           state: { activeTab: "subscribers" },
         });
       })
       .catch((err) => {
         setAlert({ type: "danger", message: err.message });
-        throw new Error(`Create or update subscriber error: ${err}`);
+        throw new Error("Create or update subscriber error: " + err);
```
lencodes: You've rewritten the back tick syntax to the `"string" + var` syntax. I'd recommend just keeping the back tick syntax here.

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -42,14 +38,22 @@ export const SubscriberForm: React.FC<SubscriberFormProps> = ({ subscriberId, en
   const [sources, setSources] = React.useState<any>(null);
   const [tableNames, setTableNames] = React.useState<Array<any>>(null);
```
lencodes: In this declaration, we're doing two things that are not how we do it in any other of our declarations:

1. We're using the `Array` keyword instead of the `[]` syntax and
2. We're specifying that `tableNames` is an array at all.

Other declarations we do like:

`const [sources, setSources] = React.useState<any>(null);`

And if we were to add that this is an array, we would usually do it like:

`const [sources, setSources] = React.useState<any[]>(null);`

For now, I'd just make the declaration like this:

`const [tableNames, setTableNames] = React.useState<any>(null);`

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -196,138 +218,85 @@ export const SubscriberForm: React.FC<SubscriberFormProps> = ({ subscriberId, en
...
-                          defaultValue={subscriber?.conditions}
+                          label={"Conditions (JSON Logic)"}
+                          tooltipContent={
+                            <a target="_blank" href="https://docs.conductor-gateway.app/en/latest/features/handlers/">
+                              Read more about the use of JSON Logic
+                            </a>
+                          }
+                          {...{ register, errors }}
+                          validation={{ validate: () => validateJSON(getValues("conditions")) }}
                         />
```
lencodes, line 253: This field has unintentionally become a `required` field due to not passing the validate when it's empty.

pwa/src/components/subscribers/subscriberForm.tsx
```diff
@@ -196,138 +218,85 @@ export const SubscriberForm: React.FC<SubscriberFormProps> = ({ subscriberId, en
...
-                                label="Translations In"
-                                data={subscriber?.translationsIn}
+                              <SelectMultiple
+                                name="translationsIn"
+                                label="Translations in"
                                 options={tableNames}
+                                {...{ control, errors }}
                               />
                             ) : (
                               <>
```
lencodes: This feedback goes for a couple of components within the accordions.

This accordion item is now set as followed:

```
{
  title: "Translations in",
  id: "translationsInAccordion",
  render: function () {
  return tableNames ? (    <== explicit return statement
    <SelectMultiple
      name="translationsIn"
      label="Translations in"
      options={tableNames}
      {...{ control, errors }}
    />
  ) : (
    <>    <== redundant wrapper
      <Spinner />
    </>
  );
},
```

I think we should update the following two aspects of this implementation (since we're refactoring the form anyway, we might as well):

1. We're using an explicit `return` statement in the `render` prop of this accordion item.  A arrow function returns its value by default if you **not not** include the curly brackets. So:

```
() => { // this does not return }
() => this does return

// and what we're doing..
() => { return // this does also return but has redundant code  }
``` 

2. If a value does not exist we want to return a `Spinner` (which is fine). However, we're wrapping this in another element, like so:

```
// current implementation
<>
  <Spinner />
< />

// a better implementation
<Spinner />
```

lencodes: Edit: for clarity, here's the final `render` prop that I think should be the implementation:


```
render: () =>                 <== no curly brackets and no return statement
  tableNames ? (
    <SelectMultiple
      name="translationsIn"
      label="Translations in"
      options={tableNames}
      {...{ control, errors }}
    />
  ) : (
    <Spinner />               <== no redundant wrapper
  ),
```

pwa/src/components/subscribers/subscriberForm.tsx
lencodes: I checked this thread https://github.com/react-hook-form/react-hook-form/issues/1781 for some answers. It seems that the only (and preferred by the creator of this package) just not validating it when there's no data available is the solution. 

I don't understand why the validation kicks in at all when it's empty and not required. That being said, we can easily fix this. 

I've left a comment on the validateJSON function with a fix 👍🏼

pwa/src/services/validateJSON.ts
```diff
@@ -1,12 +1,12 @@
...
+  if (RegExp.test(_json)) return 'Invalid JSON';
 
   try {
     JSON.parse(_json);
     return true;
   } catch {
-    return false;
+    return 'Invalid JSON';
   }
 };
```
lencodes: Couple of things here:

1. I think we can expand a bit on the error message;
2. We're not duplicating the error message;
3. We have to validate when there's no data sent (due to the way react-hook-form works) and lastly
4. We've named the prop `_json`, but we're actually validating some data to check **if** it is JSON, so it's not accurately named right now.

I propose the following:


```
export const validateJSON = (data: string): boolean | string => {
  const RegExp: RegExp = /^[0-9]+$/;
  const invalidMessage = "This field contains invalid JSON";

  if (!data) return true; // empty data should not be validated

  if (RegExp.test(data)) return invalidMessage;

  try {
    return JSON.parse(data) && true;
  } catch {
    return invalidMessage;
  }
};
```

lencodes (changes requested):
Looks good! Left a few comments.

lencodes (approved):

PR #269, opened 2022-04-01
Feature/gw 585/react query application

lencodes:
Read commit messages for contents.

### DIFF
--- a/pwa/src/hooks/application.ts
+++ b/pwa/src/hooks/application.ts
@@ -1,0 +1,83 @@
+import * as React from "react";
+import { QueryClient, useMutation, useQuery } from "react-query";
+import APIService from "../apiService/apiService";
+import APIContext from "../apiService/apiContext";
+import { AlertContext } from "../context/alertContext";
+import { LoadingOverlayContext } from "../context/loadingOverlayContext";
+import { navigate } from "gatsby-link";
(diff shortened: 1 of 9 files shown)


pwa/src/hooks/application.ts
```diff
@@ -0,0 +1,83 @@
...
+  const createOrEdit = (applicationId?: string) =>
+    useMutation<any, Error, any>(API.Application.createOrUpdate, {
+      onMutate: () => {
+        setLoadingOverlay({ isLoading: true });
+      },
+      onSuccess: async (newApplication) => {
+        if (applicationId) {
+          updateItem(queryClient, "applications", newApplication);
+          setAlert({ message: "Updated application", type: "success" });
+          navigate("/applications");
```
remko48, line 65: I think you have mixed up the navigation of create and update.
Now when we create a new application we will be redirected to the edit page of the application we just made `navigate(`/applications/${newApplication.id}`);`.
and when we update an application we will be sent back to the application overview `navigate("/applications");`.
which should be the other way around.

remko48: I also noticed that the same thing happens with handler and endpoint

remko48 (changes requested):

PR #273, opened 2022-04-02
Feature/gw 594/react query source

lencodes:
Read commit messages for contents.


remko48 (changes requested):
Everything looks good. 
Only thing is that when editing a source data from the fields: Headers, OAS and Paths won't show up in the fields. it does however get the data from the gateway.

remko48 (approved):

PR #276, opened 2022-04-05
added an input character check

remko48:
I have made small changes to the titles in some of the table files. I did this to remove the Edit part in the title.
I also made a small breadcrumb label change.

### DIFF
--- a/pwa/src/components/entities/entityForm.tsx
+++ b/pwa/src/components/entities/entityForm.tsx
@@ -11,6 +11,7 @@ import { ISelectValue } from "../formFields/types";
 import { useQueryClient } from "react-query";
 import { useEntity } from "../../hooks/entity";
 import { useSource } from "../../hooks/source";
+import { validateLength } from "../../services/validateLength";
 
 interface EntityFormProps {
   entityId: string;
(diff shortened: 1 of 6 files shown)


pwa/src/components/entities/entityForm.tsx
```diff
@@ -147,17 +149,27 @@ export const EntityForm: React.FC<EntityFormProps> = ({ entityId }) => {
                           name="function"
                           label="Function"
                           options={functionSelectOptions}
-                          validation={{ required: true }}
+                          validation={{ validate: () => validateLength(getValues("name")), required: true }}
```
lencodes: The `validation` prop can hold multiple properties. Right now, we're using `required` and you've just added `validate`.

However, you're not writing a custom validation function for a `maxLength` validation, `maxLength` is a predefined validation that `react-hook-form` offers (the same as `required`).

You can add this like: `maxLength: 255` (in case the max length is 255).

This would result in:

```
validation={{ maxLength: 255, required: true }}
```

pwa/src/services/validateLength.ts
```diff
@@ -0,0 +1,7 @@
+export const validateLength = (data: string): boolean | string => {
+  const invalidMessage = "This field contains more than the 255 character limit";
+
+  if (!data) return true; // empty data should not be validated
+
+  if (data.length > 255) return invalidMessage;
+};
```
lencodes: If you implement the `maxLength` as mentioned in the other comment, you won't need this anymore.

lencodes (changes requested):

PR #281, opened 2022-04-11
LogTable column refactor

hghanbari:


lencodes:
@hghanbari I'm unsure what happened exactly, but your branch has diverged with 251 commits from development and I can therefore not pull this to local. Let's hop in a quick call somewhere today.

lencodes:
@hghanbari I'm unsure what happened exactly, but your branch has diverged with 251 commits from development and I can therefore not pull this to local. Let's hop in a quick call somewhere today.

PR #284, opened 2022-05-24
FormIO fixes

bbrands02:


pwa/src/components/objectEntities/ObjectEntitiesTable.tsx
```diff
@@ -6,99 +6,46 @@ import APIContext from "../../apiService/apiContext";
...
-  const [formIOSchema, setFormIOSchema] = React.useState(null);
   const API: APIService = React.useContext(APIContext);
   const [_, setAlert] = React.useContext(AlertContext);
 
   React.useEffect(() => {
-    setShowSpinner(true);
     if (entityId) {
       handleSetObjectEntities();
-      getEntity();
+      checkIfEntityCanUseFormIO();
```
lencodes, line 26: More consistent name would be `checkIfEntityIsViableForFormIO`, because your state also uses _viable_.

pwa/src/components/objectEntities/ObjectEntitiesTable.tsx
```diff
@@ -6,99 +6,46 @@ import APIContext from "../../apiService/apiContext";
...
-  //     .finally(() => {
-  //       handleSetObjectEntities();
-  //     });
-  // };
-
-  const getEntity = () => {
-    setShowSpinner(true);
-    API.Entity.getOne(entityId)
+  const checkIfEntityCanUseFormIO = () => {
+    API.FormIO.getEntityCrudEndpoint(entityId)
```
lencodes, line 32: Did this end up having a delete option as well? If not, I wouldn't use the acronym CRUD.

pwa/src/components/objectEntities/ObjectEntitiesTable.tsx
```diff
@@ -146,6 +93,12 @@ const ObjectEntitiesTable: React.FC<ObjectEntitiesTableProps> = ({ entityId }) =
       cardHeader={function () {
         return (
           <>
+            {entityViableForFormIO === false && (
+              <p className="utrecht-paragraph text-left">
+                To create or edit objects, his entity needs a handler, which needs to have endpoints with GET, POST and
```
lencodes: this entity*

lencodes: which needs to have...

=>

containing GET, POST and PUT methods.

pwa/src/components/objectEntities/ObjectEntitiesTable.tsx
```diff
@@ -165,12 +118,12 @@ const ObjectEntitiesTable: React.FC<ObjectEntitiesTableProps> = ({ entityId }) =
...
-              data-bs-target="#objectModal"
-              aria-disabled="true"
-              disabled
+              disabled={!entityViableForFormIO}
+              onClick={() => {
+                navigate(`/entities/${entityId}/objects/new`);
+              }}
             >
-              <i className="fas fa-plus mr-2" />
+              <i className={"fas " + (entityViableForFormIO == null ? "fa-spinner" : "fa-plus") + " mr-2"} />
```
lencodes, line 126: No need to explicitly check if something is `null`.

`entityViableForFormIO ? "fa-spinner" : "fa-plus"` does the same as it's `...== null` counterpart.

pwa/src/components/objectEntities/ObjectEntitiesTable.tsx
```diff
@@ -180,7 +133,7 @@ const ObjectEntitiesTable: React.FC<ObjectEntitiesTableProps> = ({ entityId }) =
         return (
           <div className="row">
             <div className="col-12">
-              {showSpinner === true ? (
+              {objectEntities === false ? (
```
lencodes, line 136: Same here, no need to explicitly check if something's `false`. You can make use of `!objectEntities ? ...`

pwa/src/components/objectEntities/ObjectEntityFormNew.tsx
```diff
@@ -2,68 +2,44 @@ import * as React from "react";
...
-  objectId: string;
+  objectId?: string;
   entityId: string;
 }
 
 export const ObjectEntityFormNew: React.FC<ObjectEntityFormNewProps> = ({ objectId, entityId }) => {
   const API: APIService = React.useContext(APIContext);
-  const [object, setObject] = React.useState(null);
   const [showSpinner, setShowSpinner] = React.useState(null);
+  const [formIOEndpoint, setFormIOEndpoint] = React.useState(null);
```
lencodes, line 15: Can this be typed? 

What is its value? A _string_?

pwa/src/components/objectEntities/ObjectEntityFormNew.tsx
```diff
@@ -2,68 +2,44 @@ import * as React from "react";
...
-  React.useEffect(() => {
-    getObject();
-  }, [API]);
-
   React.useEffect(() => {
     setShowSpinner(true);
-    getEntity.isSuccess && object && getFormIOSchema();
-    setShowSpinner(false);
-  }, [getEntity.isSuccess, object]);
+    getFormIOEndpoint();
```
lencodes, line 21: Might make more sense to set spinner `true` within the `getFormIOEndpoint` method. Since we're not setting it back to `false`. 

In the current implementation, you have to search throughout the flow of this code to find the `setShowSpinner(false)`. 

I think this is an unwanted side effect.

pwa/src/components/objectEntities/ObjectEntityFormNew.tsx
```diff
@@ -73,52 +49,52 @@ export const ObjectEntityFormNew: React.FC<ObjectEntityFormNewProps> = ({ object
...
+
+    import("@formio/react").then((formio) => {
+      const { Form } = formio;
+      setFormIO(<Form src={formIOSchema} onSubmit={saveObject} />);
+    });
+  }, [formIOSchema]);
 
   const saveObject = (event) => {
+    setShowSpinner(true);
     let body = event.data;
```
lencodes, line 64: Can this be a `const`?

pwa/src/components/objectEntities/ObjectEntityFormNew.tsx
```diff
@@ -73,52 +49,52 @@ export const ObjectEntityFormNew: React.FC<ObjectEntityFormNewProps> = ({ object
...
+
     body.submit = undefined;
+    body["@application"] = undefined;
+    body["@owner"] = undefined;
+    body["@organization"] = undefined;
```
lencodes, line 69: Ideally, we would fix this within the FormIO configuration, setting these default values to `undefined` perhaps?

pwa/src/components/searchEntity/SearchEntity.css
```diff
@@ -1,7 +1,7 @@
-.SearchEntity {
+/* .SearchEntity {
   padding-block: 35px;
   padding-inline: 70px;
-}
+} */
```
lencodes, line 4: Missed a comment block.

lencodes (approved):
Please let @rubenvdlinde know that we should refactor this when we have time.

For now, approved.

Did leave some comments you might want to update.

PR #285, opened 2022-06-14
Conductor rename

MWest2020:
just changed the name from Conductor to Common Gateway.

### DIFF
--- a/pwa/src/pages/login/index.tsx
+++ b/pwa/src/pages/login/index.tsx
@@ -100,7 +100,8 @@ const Login: React.FC = () => {
       />
 
       <div className="login-container">
-        <h1>Welcome to the Common Gateway!</h1>
+        <h1>Welcome to the</h1>
+        <h1>Common Gateway!</h1>
 
         <form className="login-form" onSubmit={handleSubmit(onSubmit)}>
           <h2>Login</h2>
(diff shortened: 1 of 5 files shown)


pwa/src/pages/login/index.tsx
```diff
@@ -100,7 +100,8 @@ const Login: React.FC = () => {
       />
 
       <div className="login-container">
-        <h1>Welcome to Conductor!</h1>
+        <h1>Welcome to the</h1>
+        <h1>Common Gateway!</h1>
```
lencodes, line 104: This is semantically incorrect; a page cannot have two `h1` tags.

I expect you've added the two separate tags to create some layout. I'd recommend doing that with `css`. 

E.g.:

```
<h1 className="example">Welcome to the Common Gateway!</h1>

...

.example {
  max-width: ...;
}
```

lencodes (changes requested):
Looks good to me.

Two notes: 

1. In `gatsby-config.js` we're still setting the `gatsby-plugin-manifest` variables to "Conductor";
2. Please add ticket numbers to your branch names, e.g. feature/GW-xxx/conductor-rename

lencodes:
Note: you can just merge this whenever you've applied the feedback.

PR #286, opened 2022-06-15
added redocly link

smisidjan:


pwa/src/components/applications/applicationsTable.tsx
```diff
@@ -96,6 +96,17 @@ export default function ApplicationsTable() {
...
+                    {
+                      headerName: "Redocly",
+                      field: "name",
+                      renderCell: () => {
+                        return (
+                          <Link to={`https://redocly.github.io/redoc/?url=http://localhost/openapi.yaml&nocors`} target="__blank">
+                            See redoc
+                          </Link>
+                        );
+                      },
```
lencodes: The default functionality of an arrow function without adding curly braces (`{}`) is to return its value. Therefore, when you don't need any logic within the function body (within the `{}`) you can write this without the explicit return statement, like so:

```
renderCell: () => (
    <Link to={`https://redocly.github.io/redoc/?url=http://localhost/openapi.yaml&nocors`} target="__blank">
      See redoc
    </Link>
)
```

pwa/src/components/applications/applicationsTable.tsx
```diff
@@ -96,6 +96,17 @@ export default function ApplicationsTable() {
                       headerName: "Description",
                       field: "description",
                     },
+                    {
+                      headerName: "Redocly",
+                      field: "name",
+                      renderCell: () => {
+                        return (
+                          <Link to={`https://redocly.github.io/redoc/?url=http://localhost/openapi.yaml&nocors`} target="__blank">
+                            See redoc
```
lencodes: "View Redoc" might be a better label for this button.

Note: throughout the application we've been using "view" and not "see".

pwa/src/components/applications/applicationsTable.tsx
```diff
@@ -96,6 +96,17 @@ export default function ApplicationsTable() {
                       headerName: "Description",
                       field: "description",
                     },
+                    {
+                      headerName: "Redocly",
+                      field: "name",
+                      renderCell: () => {
+                        return (
+                          <Link to={`https://redocly.github.io/redoc/?url=http://localhost/openapi.yaml&nocors`} target="__blank">
```
lencodes: Since we're not actually doing any string interpolation we do not need the back tick syntax here (`). This syntax is meant to combine variables with strings, e.g.:

```
const complexString = `Hi there, ${name}`;
```

Also, because this prop is a string, we do not need to wrap it in curly braces (`{}`), because passing a string is valid HTML, and using JSX syntax is therefore redundant.

The result could then be:

```
<Link to="https://redocly.github.io/redoc/?url=http://localhost/openapi.yaml&nocors" target="__blank">
  See redoc
</Link>
```

lencodes: Also, I think it would be better to store this URL somewhere else, it's not content that belongs to the application table I think. Might be better placed somewhere in (e.g.) the environment variables.

lencodes (approved):
Approved, did leave some comments.

**Note:** I get `500` errors on every request, so I was unable to functionally test this.

Edit: please add issue numbers to your branches!

